PR #1, opened 2021-11-17
Initial. Simple UDP server.

HelenMamontova:
Created a simple UDP server.

### DIFF
--- a/src/server.h
+++ b/src/server.h
@@ -1,0 +1,22 @@
+#ifndef SERVER_H
+#define SERVER_H
+
+#include <boost/asio.hpp>
+#include <array>
+
+class Server
+{
+    public:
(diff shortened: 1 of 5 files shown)


src/server.h
```diff
@@ -0,0 +1,22 @@
...
+#include <array>
+
+class Server
+{
+    public:
+        Server(boost::asio::io_service& io_service);
+
+    private:
+        void start_receive();
+        void handle_receive(const boost::system::error_code& error, std::size_t /*bytes_transferred*/);
```
madf, line 14: It is not necessary to put parameter names in comments here. These are declarations.

madf: Same below.

madf (changes requested):

madf (approved):

PR #2, opened 2021-12-24
Initial. Async-radius server.

HelenMamontova:
Implemented receiving the Response Authenticator using MD5() and sending the minimum packet to the client.


src/server.cpp
```diff
@@ -20,19 +22,50 @@ void Server::start_receive()
...
-        const std::string message("Hello, World!");
+        size_t length = recv_buffer_[2] * 256 + recv_buffer_[3];
+        if (bytes >= length)
+        {
+            uint8_t send_code;
 
-        socket_.async_send_to(boost::asio::buffer(message), remote_endpoint_,
-          std::bind(&Server::handle_send, this, message,pls::_1, pls::_2));
+            if (recv_buffer_[0] == ACCESS_REQUEST)
+                send_code = 2;
```
madf, line 35: Why `2` instead of `ACCESS_ACCEPT`?

src/server.cpp
```diff
@@ -20,19 +22,50 @@ void Server::start_receive()
         std::bind(&Server::handle_receive, this, pls::_1, pls::_2));
 }
 
-void Server::handle_receive(const error_code& error, std::size_t /*bytes_transferred*/)
+void Server::handle_receive(const error_code& error, std::size_t bytes)
 {
     if (!error || error == boost::asio::error::message_size)
     {
-        const std::string message("Hello, World!");
+        size_t length = recv_buffer_[2] * 256 + recv_buffer_[3];
```
madf, line 29: What if we receive less than 4 bytes? `recv_buffer_[2]` and `recv_buffer_[3]` would be undefined.

src/server.cpp
```diff
@@ -20,19 +22,50 @@ void Server::start_receive()
         std::bind(&Server::handle_receive, this, pls::_1, pls::_2));
 }
 
-void Server::handle_receive(const error_code& error, std::size_t /*bytes_transferred*/)
+void Server::handle_receive(const error_code& error, std::size_t bytes)
 {
     if (!error || error == boost::asio::error::message_size)
```
madf, line 27: What is `boost::asio::error::message_size`? What does this error mean?

src/server.cpp
```diff
@@ -20,19 +22,50 @@ void Server::start_receive()
...
 
-        socket_.async_send_to(boost::asio::buffer(message), remote_endpoint_,
-          std::bind(&Server::handle_send, this, message,pls::_1, pls::_2));
+            if (recv_buffer_[0] == ACCESS_REQUEST)
+                send_code = 2;
 
-        start_receive();
+            std::array<uint8_t, 4096> send_buffer;
+            uint8_t send_length1 = 0;
+            uint8_t send_length2 = 20;
```
madf, line 39: What are these two variables for?

src/server.cpp
```diff
@@ -20,19 +22,50 @@ void Server::start_receive()
...
+
+            for (int i = 4; i <= 19; i++)
+                send_buffer[i] = recv_buffer_[i];
+
+            std::string secr("secret");
+
+            for (int i = 0; i <= 5; i++)
+                send_buffer[i + 20] = secr[i];
+
+            unsigned char md[MD5_DIGEST_LENGTH];
```
madf, line 54: Consider using `std::array`.

src/server.cpp
```diff
@@ -20,19 +22,50 @@ void Server::start_receive()
...
+
+            MD5(send_buffer.data(), send_length2 + secr.length(), md);
+
+            for (int i = 0; i <= 15; i++)
+                send_buffer[i + 4] = md[i];
+
+          socket_.async_send_to(boost::asio::buffer(send_buffer), remote_endpoint_,
```
madf, line 61: `send_buffer` is a local variable. When `ioc` starts actual sending, it will be well out of scope and destroyed. You need to make it permanent. Make it a class member, for example.

madf (changes requested):

src/server.cpp
```diff
@@ -20,19 +23,55 @@ void Server::start_receive()
...
+void Server::handle_receive(const error_code& error, std::size_t bytes)
 {
-    if (!error || error == boost::asio::error::message_size)
+    if (!error)
     {
-        const std::string message("Hello, World!");
+        size_t length;
+        try
+        {
+            length = recv_buffer_.at(2) * 256 + recv_buffer_.at(3);
```
madf, line 33: `recv_buffer_` is 4096 bytes long. `at` function will never throw, no reason to use it instead of `operator[]`.

The question here (which I asked in previous review) is what if we receive only 1 byte of data? Values at `recv_buffer_[2]` and `recv_buffer_[3]` would be undefined, you would get some random number in `length`.

Before accessing any received data you should check that you actually received it. In this particular case you need to check that you received at least 4 bytes of data, that 0-3 bytes of buffer are actually filled with the data.

In fact, you can easily check that you received at least 20 bytes of data - the header and the authenticator. Anything less than that would be a malformed request.

src/server.cpp
```diff
@@ -20,19 +23,55 @@ void Server::start_receive()
         std::bind(&Server::handle_receive, this, pls::_1, pls::_2));
 }
 
-void Server::handle_receive(const error_code& error, std::size_t /*bytes_transferred*/)
+void Server::handle_receive(const error_code& error, std::size_t bytes)
 {
-    if (!error || error == boost::asio::error::message_size)
+    if (!error)
```
madf, line 28: I see that in case of an error you do nothing, just exit the function. It would be nice to write something about the error, do some logging.

src/server.cpp
```diff
@@ -20,19 +23,55 @@ void Server::start_receive()
...
-          std::bind(&Server::handle_send, this, message,pls::_1, pls::_2));
+        if (bytes >= length)
+        {
+            uint8_t send_code;
 
-        start_receive();
+            if (recv_buffer_[0] == ACCESS_REQUEST)
+                send_code = ACCESS_ACCEPT;
+
+            send_buffer_[0] = send_code;
```
madf, line 47: If `recv_buffer_[0]` is not `ACCESS_REQUEST`, `send_code` would be undefined. If you don't want to handle anything other than `ACCESS_REQUEST` - just do some logging and do nothing.

madf: Also, what is the purpose of the `send_code` variable? Why can't you just assign `send_buffer_[0]` the value directly?

src/server.cpp
```diff
@@ -20,19 +23,55 @@ void Server::start_receive()
...
+            send_buffer_[1] = recv_buffer_[1];
+            send_buffer_[2] = 0;
+            send_buffer_[3] = 20;
+
+            for (int i = 4; i <= 19; i++)
+                send_buffer_[i] = recv_buffer_[i];
```
madf, line 53: I propose you to rewrite this cycle as the following:
```c++
for (size_t i = 0; i < 20; ++i)
    send_buffer_[i + 4] = recv_buffer_[i + 4];
```

The reasons are:
 * use `size_t` instead of `int` shows that you work with sizes, not some arbitrary values (and `size_t` is unsigned, a size cannot be negative);
 * start indexing from 0 to some size shows that you work with some value, not an array span;
 * it is always nice to use strict comparisons - `<` and `>` instead of `<=` and `>=`;
 * `++i` is not better that `i++` in this case, but in general you should preffer this form;
 * offset 4 shows you the position of the value you are working with in the buffer.

src/server.cpp
```diff
@@ -20,19 +23,55 @@ void Server::start_receive()
...
+            std::string secr("secret");
+
+            for (int i = 0; i <= 5; i++)
+                send_buffer_[i + 20] = secr[i];
+
+            std::array<uint8_t, 16> md;
+
+            MD5(send_buffer_.data(), 20 + secr.length(), md.data());
+
+            for (int i = 0; i <= 15; i++)
```
madf, line 64: Consider:
```c++
for (size_t i = 0; i < 16; ++i)
    send_buffer_[i + 4] = md[i];
```

src/server.cpp
madf: Besides, why 20 (19 in your case)? I think you should use 16 (15 in your version of the code).

src/server.cpp
```diff
@@ -20,19 +23,55 @@ void Server::start_receive()
...
+                send_buffer_[i + 4] = md[i];
+
+          socket_.async_send_to(boost::asio::buffer(send_buffer_), remote_endpoint_,
```
madf, line 67: I think it has indentation problem.

src/server.h
```diff
@@ -12,11 +13,12 @@ class Server
     private:
         void start_receive();
         void handle_receive(const boost::system::error_code& error, std::size_t);
```
madf, line 15: Except pure type specification, you should always use parameter names. Not just types.

src/server.h
```diff
@@ -12,11 +13,12 @@ class Server
...
-        void handle_send(std::string, const boost::system::error_code&, std::size_t);
+        void handle_send(const boost::system::error_code&, std::size_t);
```
madf, line 16: Please add parameter names.

madf (changes requested):

src/server.h
```diff
@@ -11,12 +12,13 @@ class Server
 
     private:
         void start_receive();
-        void handle_receive(const boost::system::error_code& error, std::size_t);
-        void handle_send(std::string, const boost::system::error_code&, std::size_t);
+        void handle_receive(const boost::system::error_code& error, std::size_t bytes);
+        void handle_send(const boost::system::error_code& /*error*/, std::size_t /*bytes_transferred*/);
```
madf, line 16: You don't need to put param names in `/* */` in function declarations.

src/server.cpp
```diff
@@ -20,19 +23,52 @@ void Server::start_receive()
         std::bind(&Server::handle_receive, this, pls::_1, pls::_2));
 }
 
-void Server::handle_receive(const error_code& error, std::size_t /*bytes_transferred*/)
+void Server::handle_receive(const error_code& error, std::size_t bytes)
 {
-    if (!error || error == boost::asio::error::message_size)
+    if (!error && bytes >= 20)
```
madf, line 28: You have a big nested `if` structure here. How about:
```c++
if (error)
{
    std::cout << ...; // Network error
    return;
}

if (bytes < 20)
{
    std::cout << ...; // Request size error
    return;
}

size_t length = ...;
if (btes < length)
{
    std::cout << ...; // Request size error
    return;
}

// Normal processing

```

src/server.cpp
```diff
@@ -20,19 +23,52 @@ void Server::start_receive()
...
+                send_buffer_[0] = ACCESS_ACCEPT;
+            else
+                send_buffer_[0] = ACCESS_REJECT;
 
-        start_receive();
+            send_buffer_[1] = recv_buffer_[1];
+            send_buffer_[2] = 0;
+            send_buffer_[3] = 20;
+
+            for (size_t i = 0; i < 20; ++i)
```
madf, line 43: Shouldn't we loop for 16 bytes rather than 20?

madf (changes requested):

src/server.cpp
```diff
@@ -20,19 +23,58 @@ void Server::start_receive()
...
 }
 
-void Server::handle_receive(const error_code& error, std::size_t /*bytes_transferred*/)
+void Server::handle_receive(const error_code& error, std::size_t bytes)
 {
-    if (!error || error == boost::asio::error::message_size)
+    if (error)
     {
-        const std::string message("Hello, World!");
+        std::cout << "Error async_receivw_from: " << error.message() << "/n";
```
madf, line 30: "/n" -> "\n"

madf: Same below.

src/server.cpp
```diff
@@ -20,19 +23,58 @@ void Server::start_receive()
...
+        send_buffer_[i + 20] = secr[i];
+
+    std::array<uint8_t, 16> md;
+
+    MD5(send_buffer_.data(), 20 + secr.length(), md.data());
+
+    for (size_t i = 0; i < 16; ++i)
+        send_buffer_[i + 4] = md[i];
+
+    socket_.async_send_to(boost::asio::buffer(send_buffer_), remote_endpoint_,
```
madf, line 72: You are sending the whole `send_buffer_`, but you have only 20 bytes of data in it. All other bytes are garbage!

Don't send garbage, send just the data you need.

madf (changes requested):

PR #3, opened 2022-02-18
Initial. Simple RADIUS server.

HelenMamontova:
Implemented filling 20 bytes of the server response packet and outputting the request header fields.


src/packet.h
```diff
@@ -0,0 +1,24 @@
...
+        uint8_t type() const;
+        uint8_t id() const;
+        const std::array<uint8_t, 16>& auth() const;
+        std::vector<uint8_t>& makeSendBuffer();
+
+    private:
+        uint8_t m_type;
+        uint8_t m_id;
+        std::array<uint8_t, 16> m_auth;
+        std::vector<uint8_t> m_sendBuffer;
```
madf, line 21: You don't need any send buffer here, `Packed` does not send anything.

src/packet.cpp
```diff
@@ -0,0 +1,69 @@
...
+}
+
+const std::array<uint8_t, 16>& Packet::auth() const
+{
+    return m_auth;
+}
+
+std::vector<uint8_t>& Packet::makeSendBuffer()
+{
+    m_sendBuffer.resize(20);
```
madf, line 34: `m_sendBuffer` should be a local variable. And thus, the return type must be a value, not a reference.

src/packet.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+    if (m_type == ACCESS_REQUEST)
```
madf, line 36: We should not make any dispatch here. The meaning of this method it to construct a byte array (vector) from the packet data.

madf: To clarify: this method should do everything you do in the constructor, but in reverse. E.g.:
```c++
res[0] = m_type;
res[1] = m_id;
// ...
```

src/packet.cpp
```diff
@@ -0,0 +1,69 @@
+#include "packet.h"
+#include "packet_codes.h"
+#include <string>
+#include <iostream>
+#include <openssl/md5.h>
+
+Packet::Packet(const std::array<uint8_t, 4096>& m_recvBuffer)
```
madf, line 7: You also need actual data length here.

src/server.cpp
```diff
@@ -37,44 +38,23 @@ void Server::handle_receive(const error_code& error, std::size_t bytes)
         return;
     }
 
-    size_t length = recv_buffer_[2] * 256 + recv_buffer_[3];
+    size_t length = m_recvBuffer[2] * 256 + m_recvBuffer[3];
```
madf, line 41: Length calculation and validation must be in the `Packet` constructor.

madf (changes requested):
Here is a rough view of what I want you to implement:

зображення

src/packet.cpp
```diff
@@ -0,0 +1,72 @@
...
+    sendBuffer[1] = m_id;
+
+    sendBuffer[2] = 0;
+    sendBuffer[3] = 20;
+
+    for (size_t i = 0; i < 16; ++i)
+        sendBuffer[i + 4] = m_auth[i];
+
+    for (size_t i = 0; i < 6; ++i)
+        sendBuffer[i + 20] = secret[i];
```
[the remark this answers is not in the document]
madf: Besides, instead of `6` it's better to use `secret.length()` or `secret.size()`.

src/server.cpp
```diff
@@ -37,44 +35,45 @@ void Server::handle_receive(const error_code& error, std::size_t bytes)
         return;
     }
 
-    size_t length = recv_buffer_[2] * 256 + recv_buffer_[3];
+    try
+    {
+        Packet packet = makeResponse(Packet(m_recvBuffer, bytes));
+
+        m_socket.async_send_to(boost::asio::buffer(packet.makeSendBuffer("secret"), 20),
```
madf, line 42: You don't need `20` here. The vector returned from `makeSendBuf` must be of correct size and `boost::asio::buffer` can deduce its size.

src/server.cpp
```diff
@@ -37,44 +35,45 @@ void Server::handle_receive(const error_code& error, std::size_t bytes)
...
+            m_remoteEndpoint, std::bind(&Server::handleSend, this, pls::_1, pls::_2));
 
-    if (bytes < length)
+        startReceive();
+    }
+    catch (size_t length)
     {
         std::cout << "Error: request length " << bytes << "is less than specified in the request - " << length << "\n";
```
madf, line 49: In case of exception the server stops working - you do not call `startReceive`. The server must keep working as much as possible. A single errorneous request is not a reason to give up :)

src/server.cpp
```diff
@@ -37,44 +35,45 @@ void Server::handle_receive(const error_code& error, std::size_t bytes)
...
-    for (size_t i = 0; i < 16; ++i)
-        send_buffer_[i + 4] = md[i];
+    uint8_t id = packet.id();
 
-    socket_.async_send_to(boost::asio::buffer(send_buffer_, 20), remote_endpoint_,
-        std::bind(&Server::handle_send, this, pls::_1, pls::_2));
+    std::array<uint8_t, 16> auth = packet.auth();
 
-    start_receive();
+    return Packet(type, id, auth);
```
madf, line 77: A lot of things are going on in this function. And many of them are not necessary. Consider the following:
```c++
Packet Server::makeResponse(const Packet& request)
{
    if (request.type() == ACCESS_REQUEST)
    {
        std::cout << "Packet type: ACCESS_ACCEPT\n";
        return Packet(ACCESS_ACCEPT, request.id(), request.auth());
    }
    std::cout << "Packet type: ACCESS_REJECT\n";
    return Packet(ACCESS_REJECT, request.id(), request.auth());
}
```

What did I change? First, I removed `const` from the return type. There is no reason for const when you returning a value (but a lot of reason if you returning a referemce). Second, I renamed `packet` to `request`. The word 'packet' does not tell you much, it's just a packet, and you know it already from its type. But 'request' tells you that the data in this valiable is a request. And it's type says that the request is of type 'Packet'. Which is twice as much information! Third, I removed all those useless intermediate variables. As I often say, you don't need variables unless it is absolutely necessary.

src/server.h
```diff
@@ -11,14 +12,14 @@ class Server
         Server(boost::asio::io_service& io_service);
 
     private:
-        void start_receive();
-        void handle_receive(const boost::system::error_code& error, std::size_t bytes);
-        void handle_send(const boost::system::error_code& error, std::size_t bytes_transferred);
+        void startReceive();
+        void handleReceive(const boost::system::error_code& error, std::size_t bytes);
+        void handleSend(const boost::system::error_code& error, std::size_t bytes_transferred);
+        const Packet makeResponse(const Packet& packet);
```
madf, line 18: As I mentioned above, you don't need `const` in the return type here.

src/packet.cpp
```diff
@@ -0,0 +1,73 @@
+#include "packet.h"
+#include <openssl/md5.h>
+#include <iostream>
+#include <stdexcept>
+
+Packet::Packet(const std::array<uint8_t, 4096>& m_recvBuffer, size_t bytes)
+{
+    size_t length = m_recvBuffer[2] * 256 + m_recvBuffer[3];
```
madf, line 8: It would be nice to check that `size` is at least 20 bytes (the buffer contains the whole header).

src/packet.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+    m_type = m_recvBuffer[0];
+
+    m_id = m_recvBuffer[1];
+
+    for (std::size_t i = 0; i < m_auth.size(); ++i)
+        m_auth[i] = m_recvBuffer[i + 4];
+}
+
+Packet::Packet(uint8_t type, uint8_t id, const std::array<uint8_t, 16>& auth)
```
madf, line 22: Use initialization list for `m_type` and `m_id`.

src/packet.cpp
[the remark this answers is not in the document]
madf: This problem still exists.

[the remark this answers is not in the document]
madf: Nevermind, missed you call to `resize`.

src/packet.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+    sendBuffer[1] = m_id;
+
+    sendBuffer[2] = 0;
+    sendBuffer[3] = 20;
+
+    for (size_t i = 0; i < m_auth.size(); ++i)
+        sendBuffer[i + 4] = m_auth[i];
+
+    sendBuffer.resize(26);
```
madf, line 61: Should be `sendBuffer.resize(sendBuffer.size() + secret.length())`

src/packet.cpp
```diff
@@ -0,0 +1,73 @@
...
+        sendBuffer[i + 20] = secret[i];
+
+    std::array<uint8_t, 16> md;
+
+    MD5(sendBuffer.data(), 20 + secret.length(), md.data());
+
+    for (size_t i = 0; i < md.size(); ++i)
+        sendBuffer[i + 4] = md[i];
+
+    return sendBuffer;
```
madf, line 72: There is a problem here. You are returning a buffer of 26 bytes (if you use 'secret' as secret). But actual response is only 20 bytes. Extra bytes are used in new authenticator calculation and should not be a part of the response.

madf (changes requested):

src/packet.cpp
```diff
@@ -0,0 +1,74 @@
...
+    for (std::size_t i = 0; i < m_auth.size(); ++i)
+        m_auth[i] = m_recvBuffer[i + 4];
+}
+
+Packet::Packet(uint8_t type, uint8_t id, const std::array<uint8_t, 16>& auth)
+    : m_type(type),
+      m_id(id)
+{
+    for (std::size_t i = 0; i < m_auth.size(); ++i)
+        m_auth[i] = auth[i];
```
madf, line 29: Actually, you can also use initialization list for `m_auth`.

src/packet.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+    sendBuffer[1] = m_id;
+
+    sendBuffer[2] = 0;
+    sendBuffer[3] = 20;
+
+    for (size_t i = 0; i < m_auth.size(); ++i)
+        sendBuffer[i + 4] = m_auth[i];
+
+    sendBuffer.resize(sendBuffer.size() + secret.length());
```
madf, line 61: You can allocate the correct buffer size in the constructor. No need for resize here.

madf (changes requested):

src/server.cpp
```diff
@@ -37,44 +35,33 @@ void Server::handle_receive(const error_code& error, std::size_t bytes)
...
-    for (size_t i = 0; i < 16; ++i)
-        send_buffer_[i + 4] = md[i];
-
-    socket_.async_send_to(boost::asio::buffer(send_buffer_, 20), remote_endpoint_,
-        std::bind(&Server::handle_send, this, pls::_1, pls::_2));
+        startReceive();
+    }
+    catch (const std::runtime_error& exception)
+    {
+        std::cout << "Runtime error: " << exception.what() << "\n";
```
madf, line 49: 'Runtime error' does not report any usefull information. If you have nothing to say except `exception.what()` then sey just `exception.what()`. No need to tell everybody that this is a runtime exception, it's obvious.

PR #4, opened 2022-04-27
Initial. Async-radius server.

HelenMamontova:
1. Class Attribute added.
2. Attributes type and value output implemented.


src/server.cpp
```diff
@@ -53,8 +53,64 @@ void Server::handleSend(const error_code& /*error*/, std::size_t /*bytes_transfe
 {
 }
 
+std::string typeToString(int type)
```
madf, line 56: It's better to put this function into `attribute.h/.cpp`

src/server.cpp
```diff
@@ -53,8 +53,64 @@ void Server::handleSend(const error_code& /*error*/, std::size_t /*bytes_transfe
...
+    return "unknown";
+}
+
 Packet Server::makeResponse(const Packet& request)
 {
+    std::vector<Attribute*> attributes = request.attributes();
+    for (size_t i = 0; i < attributes.size(); ++i)
+    {
+        std::cout << "Attribute type: " << typeToString(attributes[i]->type()) << "\n";
+        std::cout << "Attribute value: " << attributes[i]->value() << "\n";
```
madf, line 111: That's too verbose. Consider:
```c++
std::cout << "\t" << typeToString(attributes[i]->type()) << ": " << attributes[i]->value() << "\n";
```

src/attribute.h
```diff
@@ -0,0 +1,25 @@
...
+
+    private:
+        uint8_t m_type;
+};
+
+class String: public Attribute
+{
+    public:
+        String(uint8_t type);
+        virtual std::string value() const;
```
madf, line 22: In C++11 and later when you override a virtual method it's better to specify `override` in the end, rather than `virtual` in the beginning:
```c++
std::string value() const override;
```

src/server.cpp
```diff
@@ -53,8 +53,64 @@ void Server::handleSend(const error_code& /*error*/, std::size_t /*bytes_transfe
...
+        case 62: return "PORT_LIMIT";
+        case 63: return "LOGIN_LAT_PORT";
+    }
+    return "unknown";
+}
+
 Packet Server::makeResponse(const Packet& request)
 {
+    std::vector<Attribute*> attributes = request.attributes();
+    for (size_t i = 0; i < attributes.size(); ++i)
```
madf, line 108: In C++11 and later you can use so called 'range-for'. Like this:
```c++
for (const auto& ap : request.attributes())
{
    // 'ap' stands for 'attribute pointer'
    // ap->type()
    // ap->value()
    // ...
}
```

src/attribute.h
```diff
@@ -0,0 +1,25 @@
...
+#define ATTRIBUTE_H
+
+#include <string>
+#include <cstdint> //uint8_t, uint32_t
+
+class Attribute
+{
+    public:
+        Attribute(uint8_t type);
+        uint8_t type() const;
```
madf, line 11: How about adding:
```c++
std::string name() const;
```
With implementation like this:
```c++
std::string Attribute::name() const
{
    return typeToString(m_type);
}
```

madf (changes requested):

src/attribute.cpp
```diff
@@ -0,0 +1,72 @@
...
+std::string String::value() const
+{
+    return "1"; //temporarily
+}
+
+std::string typeToString(int type)
+{
+    switch (type)
+    {
+        case 1: return "USER_NAME";
```
madf, line 24: If you have `attribute_types.h` with names assigned to all attribute types, why do you still use numeric values?

src/server.cpp
```diff
@@ -55,6 +54,9 @@ void Server::handleSend(const error_code& /*error*/, std::size_t /*bytes_transfe
 
 Packet Server::makeResponse(const Packet& request)
 {
+    for (const auto& ap : request.attributes())
+        std::cout << "\t" << ap->name() << ": " << ap->value() << "\n";
```
madf, line 58: It would be nice to say that this list is a list of attributes. E.g.:
```
Attributes:
    USER_NAME: foobar
    FRAMED_IP_ADDRESS: 172.16.18.28
```
Not jsut a list.

Also, it is not good that you log everything in different places - packet type here, packet length in the packet constructor, request id and authenticator are not logged at all.

Let's make a special function:
```c++
void printPacket(const Packet& p);
```
You will print all information related to the packet in this function and you will call it in an appropriate place.

madf (changes requested):

src/server.cpp
```diff
@@ -8,6 +8,43 @@ using boost::system::error_code;
 
 namespace pls = std::placeholders;
 
+void printPacket(const Packet& p)
+{
+    switch (p.type())
```
madf, line 13: Let's make this `switch` a function. The same way we did for attribute type.

src/server.cpp
```diff
@@ -38,6 +75,8 @@ void Server::handleReceive(const error_code& error, std::size_t bytes)
     {
         Packet packet = makeResponse(Packet(m_recvBuffer, bytes));
 
+        printPacket(packet);
```
madf, line 78: Let's mention that this is a response packet. Just a `std::cout` before the call.

src/server.cpp
```diff
@@ -54,14 +93,10 @@ void Server::handleSend(const error_code& /*error*/, std::size_t /*bytes_transfe
 
 Packet Server::makeResponse(const Packet& request)
 {
-    for (const auto& ap : request.attributes())
-        std::cout << "\t" << ap->name() << ": " << ap->value() << "\n";
+    printPacket(request);
```
madf, line 96: And here let's say that this is a request packet.

madf (changes requested):

src/attribute.h
```diff
@@ -0,0 +1,26 @@
...
+#define ATTRIBUTE_H
+
+#include <string>
+#include <cstdint> //uint8_t, uint32_t
+
+class Attribute
+{
+    public:
+        Attribute(uint8_t type);
+        virtual ~Attribute();
```
madf, line 11: In C++11, if you are satisfied with default implementation of some methods (constructors, assignments, destructors) you can mark them `= default` to state explicitly that you want it generated by the compiler. In your case:
```c++
virtual ~Attribute() = default;
```
And that's all, you don't need the definition.

madf (changes requested):

src/server.cpp
```diff
@@ -9,6 +8,33 @@ using boost::system::error_code;
...
+        case STATUS_CLIENT: return "STATUS_CLIENT";
+    }
+    return "uncnown";
+}
+
+void printPacket(const Packet& p)
+{
+    std::cout << "Packet type: " << packetTypeToString(p.type()) << "\n";
+
+    std::cout << "ID: " << int(p.id()) << "\n";
```
madf, line 31: `int(something)` is a C-style cast, and it is considered bad. You have two options here:
```c++
std::cout << "ID: " << static_cast<unsigned>(p.id()) << "\n";
```
(`unsigned` because `uint8_t` is unsigned)

or
```c++
std::cout << "ID: " << std::to_string(p.id()) << "\n";
```
(`std::to_string` does correct conversion for numeric types).

madf (changes requested):

src/packet.cpp
```diff
@@ -24,15 +24,17 @@ Packet::Packet(const std::array<uint8_t, 4096>& m_recvBuffer, size_t bytes)
...
-        std::cout << "Attribute type: " << std::to_string(attributeType) << "\n";
-        const uint8_t attributeLength = m_recvBuffer[i + 1];
-        std::cout << "Attribute length: " << std::to_string(attributeLength) << "\n";
-        i += attributeLength;
+        const uint8_t attributeType = m_recvBuffer[attributeIndex];
+        const uint8_t attributeLength = m_recvBuffer[attributeIndex + 1];
+
+        if (attributeType == 1)
+            m_attributes.push_back(new String(attributeType, m_recvBuffer, attributeIndex));
+        if (attributeType == 5)
```
madf, line 35: We don't need to check values multiple times. Either use `switch`/`case` or `if`/`else if`:
```c++
if (attributeType == 1)
    // do something
else if (attributeType == 5)
    // do something else
```

madf (changes requested):

src/packet.cpp
```diff
@@ -20,6 +20,31 @@ Packet::Packet(const std::array<uint8_t, 4096>& m_recvBuffer, size_t bytes)
...
+    while (attributeIndex < length)
+    {
+        const uint8_t attributeType = m_recvBuffer[attributeIndex];
+        const uint8_t attributeLength = m_recvBuffer[attributeIndex + 1];
+
+        if (attributeType == 1)
+        {
+            m_attributes.push_back(new String(attributeType, &m_recvBuffer[attributeIndex + 2], attributeLength - 2));
+        }
+        else if (attributeType == 5)
```
madf, line 34: This chain of `if`s will grow quite big when you add more attribute- and data-types. And the sole purpose of these `if`s is to turn a triple (type, data pointer, data size) into an attribute pointer. Sounds like a definition of a function! So how about putting the `if`s into a function like this:
```c++
Attribute* makeAttribute(uint8_t type, const uint8_t* data, size_t size);

// ...

while (...)
{
    // ...
    try
    {
        m_attributes.push(makeAttribute(attributeType, &m_recvBuffer[attributIndex + 2], attributeLangth - 2));
    }
    catch (...)
    {
        // ...
    }
}
```

madf (changes requested):

PR #5, opened 2022-09-13
Async-radius server.

HelenMamontova:
IpAddress, Encrypted, Bytes classes added.
Method toVector() and the second constructor added to String, Integer, IpAddress, Encrypted, Bytes classes.

### DIFF
--- a/src/attribute.cpp
+++ b/src/attribute.cpp
@@ -228,7 +228,7 @@ std::string Bytes::value() const
     std::string value;
 
     for (const auto& b : m_value)
-    value += byteToHex(b);
+        value += byteToHex(b);
 
     return value;
 }
--- a/src/attribute.cpp
+++ b/src/attribute.cpp
@@ -21,7 +21,7 @@ String::String(uint8_t type, std::string name)
 {
 }
 
-std::vector<uint8_t> String::toVector(const std::string secret, std::array<uint8_t, 16> auth) const
+std::vector<uint8_t> String::toVector(const std::string& secret, std::array<uint8_t, 16> auth) const
 {
     std::vector<uint8_t> attribute(m_value.length() + 2);
     std::copy(m_value.begin(), m_value.end(), std::next(attribute.begin(), 2));
@@ -49,7 +49,7 @@ Integer::Integer(uint8_t type, uint32_t value)
 {
 }
 
-std::vector<uint8_t> Integer::toVector(const std::string secret, std::array<uint8_t, 16> auth) const
+std::vector<uint8_t> Integer::toVector(const std::string& secret, std::array<uint8_t, 16> auth) const
 {
     std::vector<uint8_t> attribute(6);
     attribute[0] = type();
@@ -76,7 +76,7 @@ IpAddress::IpAddress(uint8_t type, std::string address)
 {
 }
 
-std::vector<uint8_t> IpAddress::toVector(const std::string secret, std::array<uint8_t, 16> auth) const
+std::vector<uint8_t> IpAddress::toVector(const std::string& secret, std::array<uint8_t, 16> auth) const
 {
     std::vector<uint8_t> attribute(4);
     size_t pos = 0;
@@ -143,7 +143,7 @@ Encrypted::Encrypted(uint8_t type, std::string password)
 {
 }
 
-std::vector<uint8_t> Encrypted::toVector(const std::string secret, std::array<uint8_t, 16> auth) const
+std::vector<uint8_t> Encrypted::toVector(const std::string& secret, std::array<uint8_t, 16> auth) const
 {
     if (m_value.length() > 128)
         throw std::runtime_error("Invalid encrypted attribute size. Should be max 128 bytes, actual size is " + std::to_string(m_value.length()));
@@ -215,7 +215,7 @@ Bytes::Bytes(uint8_t type, std::string bytes)
 {
 }
 
-std::vector<uint8_t> Bytes::toVector(const std::string secret, std::array<uint8_t, 16> auth) const
+std::vector<uint8_t> Bytes::toVector(const std::string& secret, std::array<uint8_t, 16> auth) const
 {
     std::vector<uint8_t> attribute(m_value.length());
     std::copy(m_value.begin(), m_value.end(), attribute.begin());
--- a/src/attribute.cpp
+++ b/src/attribute.cpp
@@ -235,8 +235,9 @@ std::string Encrypted::value() const
 std::string Bytes::value() const
 {
     std::string value;
-    for (size_t i = 0; i < m_value.size(); ++i)
-        value += byteToHex(m_value[i]);
+
+    for (const auto& b : m_value)
+    value += byteToHex(b);
 
     return value;
 }
--- a/src/attribute.cpp
+++ b/src/attribute.cpp
@@ -190,10 +190,11 @@ std::string byteToHex(uint8_t number)
 }
 
 Bytes::Bytes(uint8_t type, const uint8_t* data, size_t size)
-        : Attribute(type)
+        : Attribute(type),
+           m_value(size)
 {
     for (size_t i = 0; i < size; ++i)
-        m_value.push_back(data[i]);
+        m_value[i] = data[i];
 }
 
 Bytes::Bytes(uint8_t type, std::vector<uint8_t> bytes)
--- a/src/attribute.cpp
+++ b/src/attribute.cpp
@@ -49,7 +49,7 @@ Integer::Integer(uint8_t type, uint32_t value)
 {
 }
 
-std::vector<uint8_t> Integer::toVector(const std::string& secret, const std::array<uint8_t, 16>& auth) const
+std::vector<uint8_t> Integer::toVector(const std::string& /*secret*/, const std::array<uint8_t, 16>& /*auth*/) const
 {
     std::vector<uint8_t> attribute(6);
     attribute[0] = type();
@@ -77,7 +77,7 @@ IpAddress::IpAddress(uint8_t type, const std::array<uint8_t, 4>& address)
 {
 }
 
-std::vector<uint8_t> IpAddress::toVector(const std::string& secret, const std::array<uint8_t, 16>& auth) const
+std::vector<uint8_t> IpAddress::toVector(const std::string& /*secret*/, const std::array<uint8_t, 16>& /*auth*/) const
 {
     std::vector<uint8_t> attribute(6);
 
@@ -180,7 +180,7 @@ Bytes::Bytes(uint8_t type, const std::vector<uint8_t>& bytes)
 {
 }
 
-std::vector<uint8_t> Bytes::toVector(const std::string& secret, const std::array<uint8_t, 16>& auth) const
+std::vector<uint8_t> Bytes::toVector(const std::string& /*secret*/, const std::array<uint8_t, 16>& /*auth*/) const
 {
     std::vector<uint8_t> attribute(m_value);
     attribute.insert(attribute.begin(), attribute.size() + 2);
--- a/src/attribute.cpp
+++ b/src/attribute.cpp
@@ -21,7 +21,7 @@ String::String(uint8_t type, const std::string& string)
 {
 }
 
-std::vector<uint8_t> String::toVector(const std::string& secret, std::array<uint8_t, 16> auth) const
+std::vector<uint8_t> String::toVector(const std::string& secret, const std::array<uint8_t, 16>& auth) const
 {
     std::vector<uint8_t> attribute(m_value.length() + 2);
     std::copy(m_value.begin(), m_value.end(), std::next(attribute.begin(), 2));
@@ -49,7 +49,7 @@ Integer::Integer(uint8_t type, uint32_t value)
 {
 }
 
-std::vector<uint8_t> Integer::toVector(const std::string& secret, std::array<uint8_t, 16> auth) const
+std::vector<uint8_t> Integer::toVector(const std::string& secret, const std::array<uint8_t, 16>& auth) const
 {
     std::vector<uint8_t> attribute(6);
     attribute[0] = type();
@@ -77,7 +77,7 @@ IpAddress::IpAddress(uint8_t type, std::array<uint8_t, 4> address)
 {
 }
 
-std::vector<uint8_t> IpAddress::toVector(const std::string& secret, std::array<uint8_t, 16> auth) const
+std::vector<uint8_t> IpAddress::toVector(const std::string& secret, const std::array<uint8_t, 16>& auth) const
 {
     std::vector<uint8_t> attribute(6);
 
@@ -133,7 +133,7 @@ Encrypted::Encrypted(uint8_t type, std::string password)
 {
 }
 
-std::vector<uint8_t> Encrypted::toVector(const std::string& secret, std::array<uint8_t, 16> auth) const
+std::vector<uint8_t> Encrypted::toVector(const std::string& secret, const std::array<uint8_t, 16>& auth) const
 {
     if (m_value.length() > 128)
         throw std::runtime_error("Invalid encrypted attribute size. Should be max 128 bytes, actual size is " + std::to_string(m_value.length()));
@@ -203,7 +203,7 @@ Bytes::Bytes(uint8_t type, const std::vector<uint8_t>& bytes)
 {
 }
 
-std::vector<uint8_t> Bytes::toVector(const std::string& secret, std::array<uint8_t, 16> auth) const
+std::vector<uint8_t> Bytes::toVector(const std::string& secret, const std::array<uint8_t, 16>& auth) const
 {
     std::vector<uint8_t> attribute(m_value);
     auto it = attribute.begin();
--- a/src/attribute.cpp
+++ b/src/attribute.cpp
@@ -167,8 +167,8 @@ std::string byteToHex(uint8_t byte)
 }
 
 Bytes::Bytes(uint8_t type, const uint8_t* data, size_t size)
-        : Attribute(type),
-           m_value(size)
+    : Attribute(type),
+      m_value(size)
 {
     for (size_t i = 0; i < size; ++i)
         m_value[i] = data[i];
--- a/src/attribute.cpp
+++ b/src/attribute.cpp
@@ -154,9 +154,6 @@ std::vector<uint8_t> Encrypted::toVector(const std::string& secret, const std::a
         for (size_t j = 0; j < md.size(); ++j)
             *it++ = plaintext[i * 16 + j] ^ md[j];
 
-        for (size_t j = 0; j < secret.length(); ++j)
-            mdBuffer[j] = secret[j];
-
         for (size_t j = 0; j < 16; ++j)
             mdBuffer[j + secret.length()] = res[i * 16 + j];
     }
--- a/src/attribute.cpp
+++ b/src/attribute.cpp
@@ -21,7 +21,7 @@ String::String(uint8_t type, std::string name)
 {
 }
 
-std::vector<uint8_t> String::toVector(std::string secret, std::array<uint8_t, 16> auth) const
+std::vector<uint8_t> String::toVector(const std::string secret, std::array<uint8_t, 16> auth) const
 {
     std::vector<uint8_t> attribute(m_value.length() + 2);
     std::copy(m_value.begin(), m_value.end(), std::next(attribute.begin(), 2));
@@ -49,7 +49,7 @@ Integer::Integer(uint8_t type, uint32_t value)
 {
 }
 
-std::vector<uint8_t> Integer::toVector(std::string secret, std::array<uint8_t, 16> auth) const
+std::vector<uint8_t> Integer::toVector(const std::string secret, std::array<uint8_t, 16> auth) const
 {
     std::vector<uint8_t> attribute(6);
     attribute[0] = type();
@@ -76,7 +76,7 @@ IpAddress::IpAddress(uint8_t type, std::string address)
 {
 }
 
-std::vector<uint8_t> IpAddress::toVector(std::string secret, std::array<uint8_t, 16> auth) const
+std::vector<uint8_t> IpAddress::toVector(const std::string secret, std::array<uint8_t, 16> auth) const
 {
     std::vector<uint8_t> attribute(4);
     size_t pos = 0;
@@ -143,7 +143,7 @@ Encrypted::Encrypted(uint8_t type, std::string password)
 {
 }
 
-std::vector<uint8_t> Encrypted::toVector(std::string secret, std::array<uint8_t, 16> auth) const
+std::vector<uint8_t> Encrypted::toVector(const std::string secret, std::array<uint8_t, 16> auth) const
 {
     if (m_value.length() > 128)
         throw std::runtime_error("Invalid encrypted attribute size. Should be max 128 bytes, actual size is " + std::to_string(m_value.length()));
@@ -215,7 +215,7 @@ Bytes::Bytes(uint8_t type, std::string bytes)
 {
 }
 
-std::vector<uint8_t> Bytes::toVector(std::string secret, std::array<uint8_t, 16> auth) const
+std::vector<uint8_t> Bytes::toVector(const std::string secret, std::array<uint8_t, 16> auth) const
 {
     std::vector<uint8_t> attribute(m_value.length());
     std::copy(m_value.begin(), m_value.end(), attribute.begin());
--- a/src/attribute.cpp
+++ b/src/attribute.cpp
@@ -148,8 +148,8 @@ std::vector<uint8_t> Encrypted::toVector(const std::string& secret, const std::a
 
     for (size_t i = 0; i < plaintext.length() / 16; ++i)
     {
-            std::array<uint8_t, 16> md;
-            MD5(mdBuffer.data(), mdBuffer.size(), md.data());
+        std::array<uint8_t, 16> md;
+        MD5(mdBuffer.data(), mdBuffer.size(), md.data());
 
         for (size_t j = 0; j < md.size(); ++j)
             *it++ = plaintext[i * 16 + j] ^ md[j];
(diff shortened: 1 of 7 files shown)


src/attribute.h
```diff
@@ -30,8 +35,43 @@ class Integer: public Attribute
...
+
+class IpAddress : public Attribute
+{
+    public:
+        IpAddress(uint8_t type, const uint8_t* data, size_t size);
+        IpAddress(uint8_t type, std::string address);
+        std::string value() const override;
+        std::vector<uint8_t> toVector(std::string secret, std::array<uint8_t, 16> auth) const override;
+    private:
+        std::string m_value;
```
madf, line 53: IP address inner representation must be either `uint32_t` or `std::array<uint8_t, 4>`.

src/attribute.h
```diff
@@ -30,8 +35,43 @@ class Integer: public Attribute
...
+
+class Bytes: public Attribute
+{
+    public:
+        Bytes(uint8_t type, const uint8_t* data, size_t size);
+        Bytes(uint8_t type, std::string bytes);
+        std::string value() const override;
+        std::vector<uint8_t> toVector(std::string secret, std::array<uint8_t, 16> auth) const override;
+    private:
+        std::string m_value;
```
madf, line 75: Here inner representation should be `std::vector<uint8_t>`.

src/attribute.cpp
```diff
@@ -27,6 +44,188 @@ Integer::Integer(uint8_t type, const uint8_t* data, size_t size)
...
+    it = attribute.insert(it, attribute.size() + 2);
+    it = attribute.insert(it, type());
+    return attribute;
+}
+
+std::string intToHex(uint8_t number)
+{
+    const std::string digits = "0123456789ABCDEF";
+    std::string hex;
+    hex.insert(0, 1, digits[number % 16]);
```
madf, line 198: Better use `hex.append(1, digits[number % 16])`. Same below.

src/attribute.cpp
```diff
@@ -27,6 +44,188 @@ Integer::Integer(uint8_t type, const uint8_t* data, size_t size)
...
+        }
+    }
+
+    auto it = attribute.begin();
+    it = attribute.insert(it, attribute.size() + 2);
+    it = attribute.insert(it, type());
+    return attribute;
+}
+
+std::string intToHex(uint8_t number)
```
madf, line 194: Better call it `byteToHex`.

src/attribute.cpp
```diff
@@ -14,6 +15,22 @@ String::String(uint8_t type, const uint8_t* data, size_t size)
...
+{
+}
+
+std::vector<uint8_t> String::toVector(std::string secret, std::array<uint8_t, 16> auth) const
+{
+    std::vector<uint8_t> attribute(m_value.length());
+    std::copy(m_value.begin(), m_value.end(), attribute.begin());
+    auto it = attribute.begin();
+    it = attribute.insert(it, m_value.length() + 2);
+    it = attribute.insert(it, type());
```
madf, line 30: How about:
```c++
std::vector<uint8_t> attribute(m_value.length() + 2);
std::copy(m_value.begin(), m_value.end(), std::next(attribute.begin(), 2));
attribute[0] = type();
attribute[1] = m_value.length() + 2;
```

src/attribute.h
```diff
@@ -11,7 +12,9 @@ class Attribute
         Attribute(uint8_t type);
         virtual ~Attribute() = default;
         std::string name() const;
+        uint8_t type() const;
         virtual std::string value() const = 0;
+        virtual std::vector<uint8_t> toVector(std::string secret, std::array<uint8_t, 16> auth) const = 0;
```
madf, line 17: `secret` must be passed by const reference: `const std::string& secret`.

madf (changes requested):

src/attribute.cpp
```diff
@@ -27,6 +43,175 @@ Integer::Integer(uint8_t type, const uint8_t* data, size_t size)
...
+
+std::vector<uint8_t> IpAddress::toVector(const std::string& secret, std::array<uint8_t, 16> auth) const
+{
+    std::vector<uint8_t> attribute;
+    for (size_t i = 0; i < m_value.size(); ++i)
+        attribute.push_back(m_value[i]);
+
+    auto it = attribute.begin();
+    it = attribute.insert(it, attribute.size() + 2);
+    it = attribute.insert(it, type());
```
madf, line 88: Why not:
```c++
std::vector<uint8_t> attribute(6);

attribute[0] = type();
attribute[1] = attribute.size();
for (size_t i = 0; i < m_value.size(); ++i)
    attribute[i + 2] = m_value[i];
```

src/attribute.cpp
```diff
@@ -27,6 +43,175 @@ Integer::Integer(uint8_t type, const uint8_t* data, size_t size)
...
+
+    return attribute;
+}
+
+Encrypted::Encrypted(uint8_t type, const uint8_t* data, size_t size, std::string secret, std::array<uint8_t, 16> auth)
+        : Attribute(type)
+{
+    if (size > 128)
```
madf, line 96: Where does this constraint come from? Why it can not be larger than 128?

src/attribute.cpp
```diff
@@ -27,6 +43,175 @@ Integer::Integer(uint8_t type, const uint8_t* data, size_t size)
...
+
+Encrypted::Encrypted(uint8_t type, std::string password)
+    : Attribute(type),
+      m_value(password)
+{
+}
+
+std::vector<uint8_t> Encrypted::toVector(const std::string& secret, std::array<uint8_t, 16> auth) const
+{
+    if (m_value.length() > 128)
```
madf, line 139: We assume that internal representation (`m_value`) is always correct. Otherwise it would not be constructed. So this check is meaningles. But see my question above regarding the size constraint.

src/attribute.cpp
```diff
@@ -27,6 +43,175 @@ Integer::Integer(uint8_t type, const uint8_t* data, size_t size)
...
+    hex.append(1, digits[number / 16]);
+    hex.append(1, digits[number % 16]);
+    return hex;
+}
+
+Bytes::Bytes(uint8_t type, const uint8_t* data, size_t size)
+        : Attribute(type)
+{
+    for (size_t i = 0; i < size; ++i)
+        m_value.push_back(data[i]);
```
madf, line 197: It is always wise to allocate vector to its size, when it is known. So, a better approach is:
```c++
Bytes::Bytes(uint8_t type, const uint8_t* data, size_t size)
    : Attribute(type),
      m_value(size)
{
    for (size_t i = 0; i < size; ++i)
        m_value[i] = data[i];
}
```

src/attribute.cpp
```diff
@@ -27,6 +43,175 @@ Integer::Integer(uint8_t type, const uint8_t* data, size_t size)
...
+}
+
+Bytes::Bytes(uint8_t type, std::vector<uint8_t> bytes)
```
madf, line 200: `const std::vector<uint8_t>&`

src/attribute.cpp
```diff
@@ -37,6 +222,27 @@ std::string Integer::value() const
     return std::to_string(m_value);
 }
 
+std::string IpAddress::value() const
+{
+    std::string value = std::to_string(m_value[0]) + "." + std::to_string(m_value[1]) + "." + std::to_string(m_value[2]) + "." + std::to_string(m_value[3]);
```
madf, line 227: No need to make a temporary variable, just return it right away:
```c++
return std::to_string(m_value[0]) + "." + ...;
```

src/attribute.cpp
```diff
@@ -37,6 +222,27 @@ std::string Integer::value() const
...
+std::string Encrypted::value() const
+{
+    return m_value;
+}
+
+std::string Bytes::value() const
+{
+    std::string value;
+    for (size_t i = 0; i < m_value.size(); ++i)
+        value += byteToHex(m_value[i]);
```
madf, line 241: You may use range-for:
```c++
for (const auto& b : m_value)
    value += bytesToHex(b);
```
or even:
```c++
return std::accumulate(m_value.begin(), m_value.end(), std::string(), byteToHex);
```

src/attribute.h
```diff
@@ -11,7 +12,9 @@ class Attribute
         Attribute(uint8_t type);
         virtual ~Attribute() = default;
         std::string name() const;
+        uint8_t type() const;
         virtual std::string value() const = 0;
+        virtual std::vector<uint8_t> toVector(const std::string& secret, std::array<uint8_t, 16> auth) const = 0;
```
madf, line 17: `const std::array<uint8_t, 16>&`

src/packet.cpp
```diff
@@ -75,37 +96,57 @@ const std::vector<Attribute*>& Packet::attributes() const
...
-    sendBuffer[3] = 20;
-
     for (size_t i = 0; i < m_auth.size(); ++i)
         sendBuffer[i + 4] = m_auth[i];
 
+    std::vector<uint8_t> attributes;
+    for (size_t i = 0; i < m_attributes.size(); ++i)
+    {
+        std::vector<uint8_t> attribute(m_attributes[i]->toVector(secret, m_auth));
+        attributes.insert(attributes.end(), attribute.begin(), attribute.end());
```
madf, line 111: A little bit better:
```c++
for (const auto& attribute : m_attributes)
{
    const auto aData = attribute.toVector(secret, m_auh);
    sendBuffer.insert(sendBuffer.end(), aData.begin(), aData.end());
}
```

src/server.cpp
```diff
@@ -87,8 +88,17 @@ Packet Server::makeResponse(const Packet& request)
     std::cout << "Request packet\n";
     printPacket(request);
 
+    std::vector<Attribute*> attributes;
+    attributes.push_back(new String(USER_NAME, "test"));
+    attributes.push_back(new Integer(NAS_PORT, 20));
+    std::array<uint8_t, 4> address {127, 104, 22, 17};
+    attributes.push_back(new IpAddress(NAS_IP_ADDRESS, address));
+    attributes.push_back(new Encrypted(USER_PASSWORD, "password123"));
+    std::vector<uint8_t> bytes {'1', '2', '3', 'a', 'b', 'c'};
```
madf, line 97: Let's add some non-printable data:
```c++
std::vector<uint8_t> bytes{0, 13, 42, 'a', 'b', 'c'};
```
Just to make sure we always treat this type as abstract data, not anything printable.

src/packet.cpp
```diff
@@ -75,37 +96,57 @@ const std::vector<Attribute*>& Packet::attributes() const
...
-    sendBuffer.resize(20);
+    sendBuffer.resize(20 + attributes.size());
     for (size_t i = 0; i < md.size(); ++i)
         sendBuffer[i + 4] = md[i];
 
     return sendBuffer;
 }
 
-Attribute* Packet::makeAttribute(uint8_t type, const uint8_t* data, size_t size)
+Attribute* Packet::makeAttribute(uint8_t type, const uint8_t* data, size_t size, std::string secret, std::array<uint8_t, 16> auth)
```
madf, line 134: `const std::string&`
`const std::array<uint8_t, 16>&`

madf (changes requested):

src/attribute.cpp
```diff
@@ -27,6 +43,172 @@ Integer::Integer(uint8_t type, const uint8_t* data, size_t size)
...
+
+Encrypted::Encrypted(uint8_t type, const uint8_t* data, size_t size, std::string secret, std::array<uint8_t, 16> auth)
+        : Attribute(type)
+{
+    if (size > 128)
+        throw std::runtime_error("Invalid encrypted attribute size. Should be max 128 bytes, actual size is " + std::to_string(size));
+
+    std::vector<uint8_t> value(size);
+
+    size_t j = 16;
```
madf, line 100: Consider the following:
Encryption/decryption diagram

This diagram illustrates encryption and decryption process. It helps to exctract key elements for the implementation - input data, cycle iteration and output.

In the decryption process, the input data is IV (initialization vector, which we put into the buffer for hashing) and ciphertext. The output is plaintext. So the implementation would look like:

```c++
std::vector<uint8_t> mdBuffer;

// Put the IV (secret + auth) into the hash buffer
// ...

for (size_t i = 0; i < ciphertext.length() / 16; ++i)
{
    std::array<uint8_t, 16> md;
    // Make hash
    MD5(mdBuffer.data(), mdBuffer.size(), md.data());
    // XOR ciphertext with hash
    for (size_t j = 0; j < 16; ++j)
        plaintext[i * 16 + j] = ciphertext[i * 16 + j] ^ md[j];
    // Put the ciphertext block (secret + block) into the hash buffer
    // ...
}
```

In the encryption process, which you have below (`toVector`), the input data is IV and plaintext. The output is ciphertext. The cycle would look like this:
```c++
std::vector<uint8_t> mdBuffer;

// Put the IV (secret + auth) into the hash buffer
// ...

for (size_t i = 0; i < plaintext.length() / 16; ++i)
{
    std::array<uint8_t, 16> md;
    // Make hash
    MD5(mdBuffer.data(), mdBuffer.size(), md.data());
    // XOR ciphertext with hash
    for (size_t j = 0; j < 16; ++j)
        ciphertext[i * 16 + j] = plaintext[i * 16 + j] ^ md[j];
    // Put the ciphertext block (secret + block) into the hash buffer
    // ...
}
```

src/attribute.cpp
```diff
@@ -27,6 +43,172 @@ Integer::Integer(uint8_t type, const uint8_t* data, size_t size)
...
+    m_value.assign(value.begin(), value.end());
+}
+
+Encrypted::Encrypted(uint8_t type, std::string password)
+    : Attribute(type),
+      m_value(password)
+{
+}
+
+std::vector<uint8_t> Encrypted::toVector(const std::string& secret, const std::array<uint8_t, 16>& auth) const
```
madf, line 136: See my comment above, for the `Encrypted::Encrypted` constructor.

src/attribute.cpp
```diff
@@ -37,6 +219,26 @@ std::string Integer::value() const
...
+{
+    return m_value;
+}
+
+std::string Bytes::value() const
+{
+    std::string value;
+
+    for (const auto& b : m_value)
+    value += byteToHex(b);
```
madf, line 237: Wrong indentation.

madf (changes requested):

src/attribute.cpp
```diff
@@ -27,6 +43,166 @@ Integer::Integer(uint8_t type, const uint8_t* data, size_t size)
...
+
+    for (size_t i = 0; i < size / 16; ++i)
+    {
+        std::array<uint8_t, 16> md;
+
+        MD5(mdBuffer.data(), mdBuffer.size(), md.data());
+        for (size_t j = 0; j < 16; ++j)
+            plaintext[i * 16 + j] = data[i * 16 + j] ^ md[j];
+
+        for (size_t j = 0; j < secret.length(); ++j)
```
madf, line 116: I think you can skip filling secret here. It is still in th buffer, MD5 does not change the source.

src/attribute.cpp
```diff
@@ -27,6 +43,166 @@ Integer::Integer(uint8_t type, const uint8_t* data, size_t size)
...
+{
+}
+
+std::vector<uint8_t> Encrypted::toVector(const std::string& secret, const std::array<uint8_t, 16>& auth) const
+{
+    std::string plaintext = m_value;
+
+    if (plaintext.length() % 16 != 0)
+    {
+        for (size_t i = 0; i < (16 - m_value.length() % 16); ++i)
```
madf, line 138: How about:
```c++
plaintext.append(16 - m_value.length() % 16, '\0');
```

src/attribute.cpp
```diff
@@ -27,6 +43,166 @@ Integer::Integer(uint8_t type, const uint8_t* data, size_t size)
...
+        for (size_t j = 0; j < secret.length(); ++j)
+            mdBuffer[j] = secret[j];
+
+        for (size_t j = 0; j < 16; ++j)
+            mdBuffer[j + secret.length()] = ciphertext[i * 16 + j];
+    }
+
+    auto it = ciphertext.begin();
+    it = ciphertext.insert(it, ciphertext.size() + 2);
+    it = ciphertext.insert(it, type());
```
madf, line 170: How about allocating `plaintext.length() + 2` bytes from the beginning, and filling the result with offset of 2?

Besides, you may use iterators instead of arithmetics:
```c++
std::vector<uint8_t> res(plaintext.length() + 2);
res[0] = type();
res[1] = res.size();
auto it = std::next(res.begin(), 2);

for (size_t i = 0; i < plaintext.length() / 16; ++i)
{
    // Calculate hash...
    
    for (size_t j = 0; j < md.size(); ++j)
        *it++ = plaintext[i * 16 + j] ^ md[j];
        
    // Fill mdBuffer...
}
```

You may also use another iterator for walking over plaintext. E.g.:
```c++
auto cit = std::next(res.begin(), 2);
auto pit = plaintext.begin();

// ...

*cit++ = *pit++ ^ md[j];

// ...
```

Although, I don't insist on this approach.

src/attribute.cpp
```diff
@@ -27,6 +43,166 @@ Integer::Integer(uint8_t type, const uint8_t* data, size_t size)
...
+}
+
+std::string byteToHex(uint8_t number)
+{
+    const std::string digits = "0123456789ABCDEF";
+    std::string hex;
+    hex.append(1, digits[number / 16]);
+    hex.append(1, digits[number % 16]);
+    return hex;
+}
```
madf, line 181: I think, this function can be made shorter:
```c++
std::string byteToHex(uint8_t byte)
{
    static const std::string digits = "0123456789ABCDEF";
    return {digits[byte / 16], digits[byte % 16]};
}
```

You may read about this syntax here: https://www.modernescpp.com/index.php/initialization

madf (changes requested):

src/attribute.cpp
```diff
@@ -27,6 +43,155 @@ Integer::Integer(uint8_t type, const uint8_t* data, size_t size)
...
+{
+    if (size > 128)
+        throw std::runtime_error("Invalid encrypted attribute size. Should be max 128 bytes, actual size is " + std::to_string(size));
+
+    std::vector<uint8_t> mdBuffer(16 + secret.length());
+
+    for (size_t i = 0; i < secret.length(); ++i)
+        mdBuffer[i] = secret[i];
+
+    for (size_t i = 0; i < 16; ++i)
```
madf, line 103: `16` -> `auth.size()`

src/attribute.cpp
```diff
@@ -27,6 +43,155 @@ Integer::Integer(uint8_t type, const uint8_t* data, size_t size)
...
+    for (size_t i = 0; i < plaintext.length() / 16; ++i)
+    {
+            std::array<uint8_t, 16> md;
+            MD5(mdBuffer.data(), mdBuffer.size(), md.data());
+
+        for (size_t j = 0; j < md.size(); ++j)
+            *it++ = plaintext[i * 16 + j] ^ md[j];
+
+        for (size_t j = 0; j < secret.length(); ++j)
+            mdBuffer[j] = secret[j];
```
madf, line 158: You don't need to fill secret more than once.

src/attribute.cpp
```diff
@@ -27,6 +43,155 @@ Integer::Integer(uint8_t type, const uint8_t* data, size_t size)
...
+
+    if (plaintext.length() % 16 != 0)
+        plaintext.append(16 - m_value.length() % 16, '\0');
+
+    std::vector<uint8_t> mdBuffer(16 + secret.length());
+
+    for (size_t i = 0; i < secret.length(); ++i)
+        mdBuffer[i] = secret[i];
+
+    for (size_t i = 0; i < 16; ++i)
```
madf, line 141: `auth.size()`

src/attribute.cpp
```diff
@@ -27,6 +43,155 @@ Integer::Integer(uint8_t type, const uint8_t* data, size_t size)
...
+Bytes::Bytes(uint8_t type, const std::vector<uint8_t>& bytes)
+    : Attribute(type),
+      m_value(bytes)
+{
+}
+
+std::vector<uint8_t> Bytes::toVector(const std::string& secret, const std::array<uint8_t, 16>& auth) const
+{
+    std::vector<uint8_t> attribute(m_value);
+    auto it = attribute.begin();
```
madf, line 189: You don't need an iterator here.
```c++
attribute.insert(attribute.begin(), attribute.size() + 2);
attribute.insert(attribute.begin(), type());
```

src/packet.cpp
```diff
@@ -75,37 +96,54 @@ const std::vector<Attribute*>& Packet::attributes() const
...
-
     for (size_t i = 0; i < m_auth.size(); ++i)
         sendBuffer[i + 4] = m_auth[i];
+    for (const auto& attribute : m_attributes)
+    {
+        const auto aData = attribute->toVector(secret, m_auth);
+        sendBuffer.insert(sendBuffer.end(), aData.begin(), aData.end());
+    }
+
+    sendBuffer[2] = 0;
```
madf, line 112: Why the third byte is zero?

src/attribute.cpp
```diff
@@ -14,6 +15,21 @@ String::String(uint8_t type, const uint8_t* data, size_t size)
 {
 }
 
+String::String(uint8_t type, const std::string& string)
+    : Attribute(type),
+      m_value(string)
+{
+}
+
+std::vector<uint8_t> String::toVector(const std::string& secret, const std::array<uint8_t, 16>& auth) const
```
madf, line 24: `secret` and `auth` are not used in this function. Keep the names, but comment them out:
```c++
std::vector<uint8_t> String::toVector(const std::string& /*secret*/, const std::array<uint8_t, 16>& /*auth*/) const
```

Same below.

madf (changes requested):

src/attribute.cpp
```diff
@@ -27,6 +43,155 @@ Integer::Integer(uint8_t type, const uint8_t* data, size_t size)
...
+
+    std::vector<uint8_t> res(plaintext.length() + 2);
+    res[0] = type();
+    res[1] = res.size();
+    auto it = std::next(res.begin(), 2);
+
+    for (size_t i = 0; i < plaintext.length() / 16; ++i)
+    {
+            std::array<uint8_t, 16> md;
+            MD5(mdBuffer.data(), mdBuffer.size(), md.data());
```
madf, line 152: Wrong indentation.

madf (changes requested):

src/attribute.cpp
```diff
@@ -27,6 +43,151 @@ Integer::Integer(uint8_t type, const uint8_t* data, size_t size)
...
+
+std::string byteToHex(uint8_t byte)
+{
+    static const std::string digits = "0123456789ABCDEF";
+    return {digits[byte / 16], digits[byte % 16]};
+}
+
+Bytes::Bytes(uint8_t type, const uint8_t* data, size_t size)
+        : Attribute(type),
+           m_value(size)
```
madf, line 171: Indentation problem.

src/packet.cpp
```diff
@@ -29,7 +29,7 @@ Packet::Packet(const std::array<uint8_t, 4096>& m_recvBuffer, size_t bytes)
 
         try
         {
-            m_attributes.push_back(makeAttribute(attributeType, &m_recvBuffer[attributeIndex + 2], attributeLength - 2));
+            m_attributes.push_back(makeAttribute(attributeType, &m_recvBuffer[attributeIndex + 2], attributeLength - 2, "secret", m_auth));
```
madf, line 32: `"secret"` must not be hard-coded. It should be a parameter.

src/packet.cpp
```diff
@@ -38,12 +38,33 @@ Packet::Packet(const std::array<uint8_t, 4096>& m_recvBuffer, size_t bytes)
 
         attributeIndex += attributeLength;
     }
+
+    bool eapMessage = false;
+    bool messageAuthenticator = false;
+    for (size_t i = 0; i < m_attributes.size(); ++i)
+    {
+        if (m_attributes[i]->type() == 79)
```
madf, line 46: `EAP_MESSAGE` rather than `79`. Same for `80` below.

src/packet.cpp
madf: Also, let's make it configurable from a command-line option.

madf (changes requested):

src/server.cpp
```diff
@@ -36,19 +36,19 @@ void printPacket(const Packet& p)
         std::cout << "\t" << ap->name() << ": " << ap->value() << "\n";
 }
 
-Server::Server(boost::asio::io_service& io_service)
+Server::Server(boost::asio::io_service& io_service, const std::string& secret)
       : m_socket(io_service, udp::endpoint(udp::v4(), 9999))
 {
-    startReceive();
+    startReceive(secret);
```
madf, line 42: If you put `secret` into a local class member (say, `m_secret`), you won't need to pass it to member functions (`startReceive` and `handleReceive`).

src/packet.cpp
```diff
@@ -38,12 +39,33 @@ Packet::Packet(const std::array<uint8_t, 4096>& m_recvBuffer, size_t bytes)
...
+    bool messageAuthenticator = false;
+    for (size_t i = 0; i < m_attributes.size(); ++i)
+    {
+        if (m_attributes[i]->type() == EAP_MESSAGE)
+            eapMessage = true;
+
+        if (m_attributes[i]->type() == MESSAGE_AUTHENTICATOR)
+            messageAuthenticator = true;
+    }
+    try
```
madf, line 53: The purpose of this construction is unclear to me:
```c++
try
{
    if (something)
        throw error(...);
}
catch (const error& e)
{
    std::cout << e.what();
}
```

You throw an exception and immediately handle it with just printing the message. If you want to just print the message, why don't you do it directly and involve exception mechanism instead?

Also, it seems to me that throwing the exception out of the constructor makes more sense. We can not construct a packet which requires 'Message-Authenticator' but does not have it. In this case, I don't understand why you handle it right away and, generaly speacking, ignore it (the `Packet` is constructed regardless).

madf (changes requested):

PR #6, opened 2022-12-02
Radius Server

HelenMamontova:
Receiving and passing the attribute VENDOR_SPECIFIC implemented.

### DIFF
--- a/src/attribute.cpp
+++ b/src/attribute.cpp
@@ -249,7 +249,7 @@ std::string ChapPassword::toString() const
     for (const auto& b : m_value)
         value += byteToHex(b);
 
-    return std::to_string(m_chapId) + value;
+    return std::to_string(m_chapId) + " " + value;
 }
 
 uint8_t ChapPassword::chapId() const
--- a/src/attribute.cpp
+++ b/src/attribute.cpp
@@ -31,7 +31,7 @@ String::String(uint8_t type, const std::string& string)
 {
 }
 
-std::string String::value() const
+std::string String::toString() const
 {
     return m_value;
 }
@@ -64,7 +64,7 @@ Integer::Integer(uint8_t type, uint32_t value)
 {
 }
 
-std::string Integer::value() const
+std::string Integer::toString() const
 {
     return std::to_string(m_value);
 }
@@ -97,7 +97,7 @@ IpAddress::IpAddress(uint8_t type, const std::array<uint8_t, 4>& address)
 {
 }
 
-std::string IpAddress::value() const
+std::string IpAddress::toString() const
 {
     return std::to_string(m_value[0]) + "." + std::to_string(m_value[1]) + "." + std::to_string(m_value[2]) + "." + std::to_string(m_value[3]);
 }
@@ -151,7 +151,7 @@ Encrypted::Encrypted(uint8_t type, const std::string& password)
 {
 }
 
-std::string Encrypted::value() const
+std::string Encrypted::toString() const
 {
     return m_value;
 }
@@ -204,7 +204,7 @@ Bytes::Bytes(uint8_t type, const std::vector<uint8_t>& bytes)
 {
 }
 
-std::string Bytes::value() const
+std::string Bytes::toString() const
 {
     std::string value;
 
@@ -242,7 +242,7 @@ ChapPassword::ChapPassword(uint8_t type, uint8_t chapId, const std::vector<uint8
 {
 }
 
-std::string ChapPassword::value() const
+std::string ChapPassword::toString() const
 {
     std::string value;
 
--- a/src/attribute.cpp
+++ b/src/attribute.cpp
@@ -1,7 +1,6 @@
 #include "packet.h"
 #include "attribute.h"
 #include "utils.h"
-#include "attribute_types.h"
 #include <openssl/md5.h>
 #include <iostream>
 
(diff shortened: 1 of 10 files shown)


src/packet.cpp
```diff
@@ -28,7 +28,10 @@ Packet::Packet(const std::array<uint8_t, 4096>& m_recvBuffer, size_t bytes, cons
         const uint8_t attributeType = m_recvBuffer[attributeIndex];
         const uint8_t attributeLength = m_recvBuffer[attributeIndex + 1];
 
-        m_attributes.push_back(makeAttribute(attributeType, &m_recvBuffer[attributeIndex + 2], attributeLength - 2, secret, m_auth));
+        if (attributeType == 26)
```
madf, line 31: Use `VENDOR_SPECIFIC` instead of `26`.

src/packet.cpp
```diff
@@ -28,7 +28,10 @@ Packet::Packet(const std::array<uint8_t, 4096>& m_recvBuffer, size_t bytes, cons
...
+            m_vendorSpecific.push_back(makeVendorSpecific(attributeType, &m_recvBuffer[attributeIndex + 2], attributeLength - 2));
```
madf, line 32: You don't need a special `makeVendorSpeific` function. You can simply do `new VendorSpecific(...)`.

src/utils.h
```diff
@@ -0,0 +1,5 @@
+#include <string>
```
madf, line 1: Add
```c++
#include <cstdint>
```
for `uint8_t`.

src/vendor_attribute.cpp
```diff
@@ -0,0 +1,77 @@
+#include "packet.h"
+#include "utils.h"
+#include "vendor_attribute.h"
+#include "attribute_types.h"
+
+VendorSpecific::VendorSpecific(uint8_t type, const uint8_t* data, size_t size)
+    : m_type(type),
+      m_value(size)
+{
+    m_vendorId = data[0] * (1 << 24)
```
madf, line 10: According to the RFC2865, high byte here must always be zero. So it might be good to check that.

src/vendor_attribute.cpp
```diff
@@ -0,0 +1,77 @@
+#include "packet.h"
+#include "utils.h"
+#include "vendor_attribute.h"
+#include "attribute_types.h"
+
+VendorSpecific::VendorSpecific(uint8_t type, const uint8_t* data, size_t size)
+    : m_type(type),
+      m_value(size)
```
madf, line 8: Why do you initialize `m_value` to `size` if you resize it right away?

src/vendor_attribute.h
```diff
@@ -0,0 +1,25 @@
...
+class VendorSpecific
+{
+    public:
+        VendorSpecific(uint8_t type, const uint8_t* data, size_t size);
+        VendorSpecific(uint8_t type, uint32_t vendorId, uint8_t vendorType, const std::vector<uint8_t>& vendorValue);
+        std::string name() const;
+        uint8_t type() const;
+        std::string value() const;
+        uint8_t vendorType() const;
+        std::string vendorId() const;
```
madf, line 17: Why `vendorId` is `std::string`?

src/vendor_attribute.h
```diff
@@ -0,0 +1,25 @@
...
+        std::vector<uint8_t> toVector() const;
+    private:
+        uint8_t m_type;
```
madf, line 20: `type` is always `26`, `VENDOR_SPECIFIC`. Why do you need this method (and the member that stores the value)?

madf (changes requested):

src/packet.cpp
```diff
@@ -47,11 +50,12 @@ Packet::Packet(const std::array<uint8_t, 4096>& m_recvBuffer, size_t bytes, cons
         throw std::runtime_error{"The EAP-Message attribute is present, but the Message-Authenticator attribute is missing"};
 }
 
-Packet::Packet(uint8_t type, uint8_t id, const std::array<uint8_t, 16>& auth, const std::vector<Attribute*>& attributes)
+Packet::Packet(uint8_t type, uint8_t id, const std::array<uint8_t, 16>& auth, const std::vector<Attribute*>& attributes, const std::vector <VendorSpecific*>& vendorSpecific)
```
madf, line 53: Extra whitespace between `vector` and `<`

src/packet.h
```diff
@@ -11,12 +12,13 @@ class Packet
 {
     public:
         Packet(const std::array<uint8_t, 4096>& m_recvBuffer, size_t bytes, const std::string& secret);
-        Packet(uint8_t type, uint8_t id, const std::array<uint8_t, 16>& auth, const std::vector<Attribute*>& attributes);
+        Packet(uint8_t type, uint8_t id, const std::array<uint8_t, 16>& auth, const std::vector<Attribute*>& attributes, const std::vector <VendorSpecific*>& vendorSpecific);
```
madf, line 15: Extra whitespace between `vector` and `<`

madf (changes requested):

src/attribute.cpp
```diff
@@ -188,6 +182,35 @@ std::vector<uint8_t> Bytes::toVector(const std::string& /*secret*/, const std::a
...
+    : Attribute(type),
+      m_chapId(chapId),
+      m_value(chapValue)
+{
+}
+
+std::vector<uint8_t> ChapPassword::toVector(const std::string& /*secret*/, const std::array<uint8_t, 16>& /*auth*/) const
+{
+    std::vector<uint8_t> attribute(m_value);
+    attribute.insert(attribute.begin(), chapId() % 256);
```
madf, line 208: Why `chapId()` rather than `m_chapId`?

Also, why `% 256`, this field is `uint8_t`, it never exceeds 255.

src/attribute.cpp
```diff
@@ -218,55 +241,19 @@ std::string Bytes::value() const
...
-        case EAP_MESSAGE: return "EAP_MESSAGE";
-        case MESSAGE_AUTHENTICATOR: return "MESSAGE_AUTHENTICATOR";
-    }
-    return "unknown";
+    std::string value;
+
+    for (const auto& b : m_value)
+        value += byteToHex(b);
+
+    return value;
```
madf, line 251: Okay, the result contains the value. But how about the CHAP Id?

src/attribute.h
```diff
@@ -74,4 +74,17 @@ class Bytes: public Attribute
...
         std::vector<uint8_t> m_value;
 };
+
+class ChapPassword: public Attribute
+{
+    public:
+        ChapPassword(uint8_t type, const uint8_t* data, size_t size);
+        ChapPassword(uint8_t type, uint8_t chapId, const std::vector<uint8_t>& chapValue);
+        std::string value() const override;
+        uint8_t chapId() const;
```
madf, line 84: Let's add `std::vector<uint8_t> chapValue() const` which will return `m_value`.

madf (changes requested):

src/attribute.cpp
```diff
@@ -188,93 +222,51 @@ std::vector<uint8_t> Bytes::toVector(const std::string& /*secret*/, const std::a
...
-std::string Bytes::value() const
+std::string ChapPassword::toString() const
 {
     std::string value;
 
     for (const auto& b : m_value)
         value += byteToHex(b);
 
-    return value;
+    return std::to_string(m_chapId) + value;
```
madf, line 252: This way CHAP id will be hard to distinguish from the value. Let's separate them by something. Spacebar, coma, etc.

madf (changes requested):

PR #8, opened 2023-02-10
Async Radius.

HelenMamontova:
Files dictionaries.h and dictionaries.cpp with classes for filling dictionaries added.

### DIFF
--- a/src/dictionaries.cpp
+++ b/src/dictionaries.cpp
@@ -17,8 +17,8 @@ uint8_t BasicDictionary::getType(const std::string& name) const
 
 void BasicDictionary::add(uint8_t type, const std::string& name)
 {
-    m_rightDict.insert(std::make_pair(type, name));
-    m_reverseDict.insert(std::make_pair(name, type));
+    m_rightDict.emplace(std::make_pair(type, name));
+    m_reverseDict.emplace(std::make_pair(name, type));
 }
 
 DependentDictionary::DependentDictionary()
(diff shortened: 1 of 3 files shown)


src/dictionaries.cpp
```diff
@@ -0,0 +1,42 @@
...
+}
+
+uint8_t BasicDictionary::getType(const std::string& name) const
+{
+    return m_reverseDict.at(name);
+}
+
+void BasicDictionary::add(uint8_t type, const std::string& name)
+{
+    m_rightDict.insert(std::pair<uint8_t, std::string>(type, name));
```
madf, line 20: You can use `std::map::emplace`. E.g. `m_rightDict.emplace(type, name);`

src/dictionaries.cpp
```diff
@@ -0,0 +1,42 @@
...
+}
+
+uint8_t DependentDictionary::getType(const std::string& attributeName, const std::string& name) const
+{
+    return m_reverseDict.at(make_pair(attributeName, name));
+}
+
+void DependentDictionary::add(uint8_t type, const std::string& name, const std::string& attributeName)
+{
+    m_rightDict.insert(std::make_pair(std::pair<std::string, uint8_t>(attributeName, type), name));
```
madf, line 40: Why do you mix `std::make_pair` with `std::pair<...>`? If you can use `std::make_pair` - use it everywhere. No need to write pair types manually.

madf (changes requested):

src/dictionaries.cpp
```diff
@@ -0,0 +1,42 @@
...
+}
+
+uint8_t BasicDictionary::getType(const std::string& name) const
+{
+    return m_reverseDict.at(name);
+}
+
+void BasicDictionary::add(uint8_t type, const std::string& name)
+{
+    m_rightDict.emplace(std::make_pair(type, name));
```
madf, line 20: With `std::map::emplace` you don't need to construct pairs manually:
`m_rightDict.emplace(type, name);`

src/dictionaries.cpp
```diff
@@ -0,0 +1,42 @@
...
+}
+
+uint8_t DependentDictionary::getType(const std::string& attributeName, const std::string& name) const
+{
+    return m_reverseDict.at(std::make_pair(attributeName, name));
+}
+
+void DependentDictionary::add(uint8_t type, const std::string& name, const std::string& attributeName)
+{
+    m_rightDict.emplace(std::make_pair(std::make_pair(attributeName, type), name));
```
madf, line 40: Same as above, you don't need the outer `std::make_pair`.

madf (changes requested):

madf (approved):

PR #9, opened 2023-02-24
Async Radius server

HelenMamontova:
Dictionary classes implemented.

### DIFF
--- a/src/dictionaries.h
+++ b/src/dictionaries.h
@@ -33,11 +33,11 @@ class Dictionaries
 {
     public:
         Dictionaries();
-        const BasicDictionary& getAttributes();
-        const BasicDictionary& getVendorNames();
-        const DependentDictionary& getAttributeValues();
-        const DependentDictionary& getVendorAttributes();
-        const DependentDictionary& getVendorAttributeValues();
+        const BasicDictionary& getAttributes() const;
+        const BasicDictionary& getVendorNames() const;
+        const DependentDictionary& getAttributeValues() const;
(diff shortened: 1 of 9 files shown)


src/dictionaries.h
```diff
@@ -29,4 +29,20 @@ class DependentDictionary
         std::map<std::pair<std::string, std::string>, uint8_t> m_reverseDict;
 };
 
+class Dictionaries
+{
+    public:
+        Dictionaries();
```
madf, line 35: In C++11 and later (here we use C++17) if you have an emty constructor you can mark it as `default` to ask the compiler to generate it for you. So instead of having
```c++
Dictionaries();
```
in the header and
```c++
Dictionaries::Dictionaries()
{
}
```
in the cpp-file, you can write:
```c++
Dictionaries() = default;
```
in the header and nothing in the cpp-file.

src/dictionaries.h
```diff
@@ -29,4 +29,20 @@ class DependentDictionary
...
+        const BasicDictionary& getAttributes() const;
```
madf, line 36: I think it would be better not to use `get` prefix in these methods. Usually, `get` implies `set`, but we do not have setters. Also, we are not getting anything here, we simpli refer some values. So `attributes` instead of `getAttributes`, `vendoerNames` instead of `getVendorNames`, and so on.

Also, since you simply return private members in these methods, you can define them inline:
```c++
const BasicDictionary& attributes() const { return m_attributes; }
```
No need to pollute cpp-file with one-liners.

madf (changes requested):

PR #10, opened 2023-03-07
Async-Radius server

HelenMamontova:
Read dictionary file Radius added.

### DIFF
--- a/src/dictionaries.cpp
+++ b/src/dictionaries.cpp
@@ -35,7 +35,7 @@ void DependentDictionary::add(uint8_t type, const std::string& name, const std::
     m_reverseDict.emplace(std::make_pair(attributeName, name), type);
 }
 
-Dictionaries::Dictionaries(const std::string& filePath)
+Dictionaries::Dictionaries(const std::string& filePath, const std::string& keyword)
 {
     std::ifstream stream(filePath);
     if (!stream)
@@ -43,5 +43,22 @@ Dictionaries::Dictionaries(const std::string& filePath)
 
     std::string line;
     while (std::getline(stream, line))
-        std::cout << line << "\n";
+    {
+        size_t firstPosKeyword = line.find(keyword);
+        if (firstPosKeyword != line.npos)
+        {
+            size_t lastPosKeyword = line.find_first_of(" \t", firstPosKeyword) - 1;
+
+            size_t firstPosName = line.find_first_not_of(" \t", lastPosKeyword + 1);
+            size_t lastPosName = line.find_first_of("  \t", firstPosName) - 1;
+            std::string name = line.substr(firstPosName, lastPosName - firstPosName + 1);
+
+            size_t firstPosType = line.find_first_not_of(" \t", lastPosName + 1);
+            size_t lastPosType= line.find_first_of("  \t", firstPosType) - 1;
+            std::string type = line.substr(firstPosType, lastPosType - firstPosType + 1);
+
+            if (name == "User-Name" || name == "User-Password")
+                std::cout << name << ": " << type << "\n";
+        }
+    }
 }
--- a/src/dictionaries.cpp
+++ b/src/dictionaries.cpp
@@ -43,6 +43,7 @@ Dictionaries::Dictionaries(const std::string& filePath)
 
     std::string line;
     const std::string keyword("ATTRIBUTE");
+    size_t lineNumber = 0;
     while (std::getline(stream, line))
     {
         size_t firstPosKeyword = line.find(keyword, 0);
@@ -53,8 +54,9 @@ Dictionaries::Dictionaries(const std::string& filePath)
         else if (firstPosKeyword != std::string::npos)
         {
             size_t firstPosName = line.find_first_not_of(" \t", firstPosKeyword + keyword.size());
+
             if (firstPosName == std::string::npos)
-                throw std::runtime_error("Attribute name not found.");
+                throw std::runtime_error(filePath + ":" + std::to_string(lineNumber) + ": Attribute name is missing in the attribute definition.");
 
             std::string name = line.substr(firstPosName, line.find_first_of("  \t", firstPosName) - firstPosName);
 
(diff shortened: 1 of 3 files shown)


src/dictionaries.cpp
```diff
@@ -32,3 +34,17 @@ void DependentDictionary::add(uint8_t type, const std::string& name, const std::
     m_rightDict.emplace(std::make_pair(attributeName, type), name);
     m_reverseDict.emplace(std::make_pair(attributeName, name), type);
 }
+
+Dictionaries::Dictionaries(const std::string& filePath)
+{
+    std::ifstream inf(filePath);
+    if (!inf)
+        throw std::runtime_error("File Dictonary " + filePath + " failed to open.");
```
madf, line 42: `"Cannot open dictionary file " + filePath`

or

`"Failed to open dictionary file" + filePath`

src/dictionaries.cpp
```diff
@@ -32,3 +34,17 @@ void DependentDictionary::add(uint8_t type, const std::string& name, const std::
...
+
+    while (inf)
```
madf, line 44: There is a common way to use `std::getline` in a cycle. You can do either
```c++
std::string line;
while (std::getline(stream, line))
{
    // ...
}
```
or
```c++
for (std::string line; std::getline(stream, line);)
{
    // ...
}
```

madf (changes requested):

src/dictionaries.cpp
```diff
@@ -32,3 +34,31 @@ void DependentDictionary::add(uint8_t type, const std::string& name, const std::
     m_rightDict.emplace(std::make_pair(attributeName, type), name);
     m_reverseDict.emplace(std::make_pair(attributeName, name), type);
 }
+
+Dictionaries::Dictionaries(const std::string& filePath, const std::string& keyword)
```
madf, line 38: What is `keyword`? Why do you need it?

src/dictionaries.cpp
```diff
@@ -32,3 +34,31 @@ void DependentDictionary::add(uint8_t type, const std::string& name, const std::
...
+{
+    std::ifstream stream(filePath);
+    if (!stream)
+        throw std::runtime_error("Cannot open dictionary file " + filePath);
+
+    std::string line;
+    while (std::getline(stream, line))
+    {
+        size_t firstPosKeyword = line.find(keyword);
+        if (firstPosKeyword != line.npos)
```
madf, line 48: Not `line.npos`. Use `std::string::npos` instead.

src/dictionaries.cpp
```diff
@@ -32,3 +34,31 @@ void DependentDictionary::add(uint8_t type, const std::string& name, const std::
...
+        {
+            size_t lastPosKeyword = line.find_first_of(" \t", firstPosKeyword) - 1;
+
+            size_t firstPosName = line.find_first_not_of(" \t", lastPosKeyword + 1);
+            size_t lastPosName = line.find_first_of("  \t", firstPosName) - 1;
+            std::string name = line.substr(firstPosName, lastPosName - firstPosName + 1);
+
+            size_t firstPosType = line.find_first_not_of(" \t", lastPosName + 1);
+            size_t lastPosType= line.find_first_of("  \t", firstPosType) - 1;
+            std::string type = line.substr(firstPosType, lastPosType - firstPosType + 1);
```
madf, line 58: Let's call it `code` or `oid` (from the https://freeradius.org/radiusd/man/dictionary.html) or `number` (from the https://networkradius.com/doc/current/concepts/dictionary/format.html)

madf (changes requested):

src/dictionaries.cpp
```diff
@@ -1,34 +1,70 @@
...
+        throw std::runtime_error("Cannot open dictionary file " + filePath);
+
+    std::string line;
+    const std::string keyword("ATTRIBUTE");
+    while (std::getline(stream, line))
+    {
+        size_t firstPosKeyword = line.find(keyword, 0);
+        if (firstPosKeyword == std::string::npos && stream.eof())
+        {
+            throw std::runtime_error("Keyword " + keyword + " not found.");
```
madf, line 51: Якщо ми не знайшли ключево слово ATTRIBUTE (`firstPosKeywork == std::string::npos`) і закінчили читати файл (`stream.eof()`) то це не є проблемою. Я на бачу причини викидувати exception.

src/dictionaries.cpp
```diff
@@ -1,34 +1,70 @@
...
+        }
+        else if (firstPosKeyword != std::string::npos)
+        {
+            size_t firstPosName = line.find_first_not_of(" \t", firstPosKeyword + keyword.size());
+            if (firstPosName == std::string::npos)
+                throw std::runtime_error("Attribute name not found.");
```
madf, line 57: Your error message here is correct, but too generic. Imagine a user feeding a file to the server. The server says "Attribute name not found." In what file? At what line? There is nothing about that in the message. So you can't figure out which file is corrupted and where exactly. I propose to add a file name and a line number to the message. Something like this: `<file-name>:<line-number> Attribute name is missing in the attribute definition.`

src/dictionaries.cpp
```diff
@@ -1,34 +1,70 @@
...
+
+            std::string name = line.substr(firstPosName, line.find_first_of("  \t", firstPosName) - firstPosName);
+
+            size_t firstPosType = line.find_first_not_of(" \t", firstPosName + name.size());
```
madf, line 61: We agreed to use `code` instead of `type`.

src/dictionaries.cpp
```diff
@@ -1,34 +1,70 @@
...
+            if (firstPosType == std::string::npos)
+                throw std::runtime_error("Attribute code not found.");
```
madf, line 63: Same as in the missing name above. However the error message can be even more informative here, because we have attribute name! Something like: `<file-name>:<line-number> Attribute code is missing in the attribute '<name>' definition.`

madf (changes requested):

madf (approved):

PR #11, opened 2023-04-03
Simple RADIUS server

HelenMamontova:
Output of names and codes of attributes and vendor attributes from file dictionary implemented.

### DIFF
--- a/src/dictionaries.cpp
+++ b/src/dictionaries.cpp
@@ -57,26 +57,26 @@ Dictionaries::Dictionaries(const std::string& filePath)
         for (const auto &t : tok)
             tokens.push_back(t);
 
-        for (size_t i = 0; i < tokens.size(); ++i)
+        if (tokens.size() != 0)
         {
-            if (tokens[i] == "ATTRIBUTE")
+            if (tokens[0] == "ATTRIBUTE")
             {
-                std::string name = tokens[i + 1];
-                std::string code = tokens[i + 2];
+                std::string name = tokens[1];
+                std::string code = tokens[2];
                 if (vendorFlag)
                     std::cout << "  " << name << ": " << code << "\n";
                 else
                     std::cout << name << ": " << code << "\n";
             }
-            else if (tokens[i] == "VENDOR")
+            else if (tokens[0] == "VENDOR")
             {
-                std::string vendorName = tokens[i + 1];
-                std::string vendorCode = tokens[i + 2];
+                std::string vendorName = tokens[1];
+                std::string vendorCode = tokens[2];
                 std::cout << "Vendor " << vendorName << ": " << vendorCode << "\n";
             }
-            else if (tokens[i] == "BEGIN-VENDOR")
+            else if (tokens[0] == "BEGIN-VENDOR")
                 vendorFlag = true;
-            else if (tokens[i] == "END-VENDOR")
+            else if (tokens[0] == "END-VENDOR")
                 vendorFlag = false;
         }
         ++lineNumber;
--- a/src/dictionaries.cpp
+++ b/src/dictionaries.cpp
@@ -44,7 +44,7 @@ Dictionaries::Dictionaries(const std::string& filePath)
         throw std::runtime_error("Cannot open dictionary file " + filePath);
 
     using tokenizer =  boost::tokenizer<boost::char_separator<char>>;
-    boost::char_separator<char> sep("\t");
+    boost::char_separator<char> sep(" \t");
 
     std::string line;
     size_t lineNumber = 0;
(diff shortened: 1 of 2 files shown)


src/dictionaries.cpp
```diff
@@ -41,28 +43,41 @@ Dictionaries::Dictionaries(const std::string& filePath)
     if (!stream)
         throw std::runtime_error("Cannot open dictionary file " + filePath);
 
+    typedef boost::tokenizer<boost::char_separator<char>> tokenizer;
```
madf, line 46: In C++11 and later versions we do not use `typedef`. We use `using` instead:
```c++
using tokenizer = boost::tokenizer<boost::char_separator<char>>;
```

src/dictionaries.cpp
```diff
@@ -41,28 +43,41 @@ Dictionaries::Dictionaries(const std::string& filePath)
...
+            tokens.push_back(t);
 
-            size_t firstPosCode = line.find_first_not_of(" \t", firstPosName + name.size());
-            if (firstPosCode == std::string::npos)
-                throw std::runtime_error(filePath + ":" + std::to_string(lineNumber) + ": Attribute code is missing in the attribute " + name + " definition.");
-
-            std::string code = line.substr(firstPosCode, line.find_first_of("  \t", firstPosCode) - firstPosCode);
-
-            std::cout << name << ": " << code << "\n";
+        for (size_t i = 0; i < tokens.size(); ++i)
```
madf, line 60: What is the purpose of this cycle?

madf (changes requested):

src/dictionaries.cpp
```diff
@@ -41,28 +43,41 @@ Dictionaries::Dictionaries(const std::string& filePath)
     if (!stream)
         throw std::runtime_error("Cannot open dictionary file " + filePath);
 
+    using tokenizer =  boost::tokenizer<boost::char_separator<char>>;
+    boost::char_separator<char> sep("\t");
```
madf, line 47: I read the doc carefully and also I did some experiments. First, from the doc, there are two constructors for `boost_char_separator`. One with arguments and one default (no arguments). The default one creates a separator which use `std::isspace` for droppped delimeters and `std::ispunct` for kept delimeters. It means that this constructor will tokenize by space-like characters (whitespace, tabulation, etc.) and punctuation (coma, semicolon, etc.). It will drop the whitespaces but keep the punctuation. The following code:
```c++
int main()
{
  std::string str = "Foo , bar,    baz,     huzzy";
  using tokenizer = boost::tokenizer<boost::char_separator<char>>;
  boost::char_separator<char> sep;
  tokenizer tokens(str, sep);
  for (const auto& tok : tokens)
      std::cout << "<" << tok << "> ";
  std::cout << "\n";
  return EXIT_SUCCESS;
}
```
will result in the following output:
```
<Foo> <,> <bar> <,> <baz> <,> <huzzy>
```
(notice that the words do not contain spaces, but punctuation is kept).

The other constructor have 3 arguments: dropped delimiters, kept delimiters and empty token policy. First argument contains a list of delimiters that will be dropped (like whitespaces in the previous example). Second argument contains a list of delimiters that will be kept (like punktuation in the example above). Third argument lets you decide whether to keep empty tokens (e.g., the string ",," has 3 empty tokens if we use ',' as a delimiter). By default, the second argument is an empty string (no kept tokens) and the third is configured to drop empty tokens. So you only need to specify the first argument. Here is an example:
```c++
int main()
{
  std::string str = "Foo , bar,    baz,     huzzy";
  using tokenizer = boost::tokenizer<boost::char_separator<char>>;
  boost::char_separator<char> sep(",");
  tokenizer tokens(str, sep);
  for (const auto& tok : tokens)
      std::cout << "<" << tok << "> ";
  std::cout << "\n";
  return EXIT_SUCCESS;
}
```
The output is:
```
<Foo > < bar> <    baz> <     huzzy>
```
See how the words were separated by comas, but the whitespaces are kept in place.

To drop whitespaces you need to specify them in the delimiter list:
```c++
int main()
{
  std::string str = "Foo , bar,    baz,     huzzy";
  using tokenizer = boost::tokenizer<boost::char_separator<char>>;
  boost::char_separator<char> sep(", ");
  tokenizer tokens(str, sep);
  for (const auto& tok : tokens)
      std::cout << "<" << tok << "> ";
  std::cout << "\n";
  return EXIT_SUCCESS;
}
```
The result is a nice list of words without any whitespaces and punctuation:
```
<Foo> <bar> <baz> <huzzy>
```

So in your case if you want to split line by whitespaces you may either use the default constructor (like in the first example) or specify all space-like symbols, not only the tabulation.

madf (changes requested):

PR #12, opened 2023-04-12
Simple RADIUS server.

HelenMamontova:
Implemented filling dictionaries and output from them.

### DIFF
--- a/src/dictionaries.cpp
+++ b/src/dictionaries.cpp
@@ -63,7 +63,7 @@ Dictionaries::Dictionaries(const std::string& filePath)
             if (tokens[0] == "ATTRIBUTE")
             {
                 const auto& attrName = tokens[1];
-                const auto& code = std::stoul(tokens[2]);
+                const auto code = std::stoul(tokens[2]);
                 if (!vendorName.empty())
                     m_vendorAttributes.add(code, attrName, vendorName);
                 else
@@ -73,7 +73,7 @@ Dictionaries::Dictionaries(const std::string& filePath)
             {
                 const auto& attrNameVal = tokens[1];
                 const auto& valueName = tokens[2];
-                const auto& valueCode = std::stoul(tokens[3]);
+                const auto valueCode = std::stoul(tokens[3]);
                 if (!vendorName.empty())
                     m_vendorAttributeValues.add(valueCode, valueName, attrNameVal);
                 else
--- a/src/dictionaries.cpp
+++ b/src/dictionaries.cpp
@@ -63,21 +63,21 @@ Dictionaries::Dictionaries(const std::string& filePath)
             if (tokens[0] == "ATTRIBUTE")
             {
                 const auto& attrName = tokens[1];
-                const auto& code = tokens[2];
+                const auto& code = std::stoul(tokens[2]);
                 if (vendorName.size() != 0)
-                    m_vendorAttributes.add(std::stoul(code), attrName, vendorName);
+                    m_vendorAttributes.add(code, attrName, vendorName);
                 else
-                    m_attributes.add(std::stoul(code), attrName);
+                    m_attributes.add(code, attrName);
             }
             else if (tokens[0] == "VALUE")
             {
                 const auto& attrNameVal = tokens[1];
                 const auto& valueName = tokens[2];
-                const auto& valueCode = tokens[3];
+                const auto& valueCode = std::stoul(tokens[3]);
                 if (vendorName.size() != 0)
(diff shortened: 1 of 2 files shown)


src/dictionaries.cpp
```diff
@@ -5,33 +5,33 @@
...
 
-void BasicDictionary::add(uint8_t code, const std::string& name)
+void BasicDictionary::add(uint32_t code, const std::string& name)
 {
     m_rightDict.emplace(code, name);
     m_reverseDict.emplace(name, code);
 }
 
-std::string DependentDictionary::name(const std::string& attributeName, uint8_t code) const
+std::string DependentDictionary::name(const std::string& attributeName, uint32_t code) const
```
madf, line 24: I thought we agreed to use 'dependencyName' instead of 'attributeName' here.

src/dictionaries.cpp
```diff
@@ -49,6 +49,14 @@ Dictionaries::Dictionaries(const std::string& filePath)
     std::string line;
     size_t lineNumber = 0;
     bool vendorFlag = false;
+
+    BasicDictionary attr;
+    DependentDictionary attrValue;
+    BasicDictionary vendName;
+    DependentDictionary vendAttr;
+    DependentDictionary vendAttrValue;
```
madf, line 57: What are these? Why don't you use `m_attribues`, `m_vendorNames`, etc.?

src/dictionaries.cpp
```diff
@@ -49,6 +49,14 @@ Dictionaries::Dictionaries(const std::string& filePath)
     std::string line;
     size_t lineNumber = 0;
     bool vendorFlag = false;
```
madf, line 51: Do you still need `vendorFlag` if you have `vendorName`? I mean, you can replace `vendorFlag` check by seeing if the `vendorName` is empty or not.

src/dictionaries.cpp
```diff
@@ -61,18 +69,29 @@ Dictionaries::Dictionaries(const std::string& filePath)
         {
             if (tokens[0] == "ATTRIBUTE")
             {
-                std::string name = tokens[1];
+                std::string attrName = tokens[1];
                 std::string code = tokens[2];
                 if (vendorFlag)
-                    std::cout << "  " << name << ": " << code << "\n";
+                    vendAttr.add(std::stoi(code), attrName, vendorName);
```
madf, line 75: How about using `std::stoul`? We don't allow negative code values, do we?

src/dictionaries.cpp
```diff
@@ -61,18 +69,29 @@ Dictionaries::Dictionaries(const std::string& filePath)
         {
             if (tokens[0] == "ATTRIBUTE")
             {
-                std::string name = tokens[1];
+                std::string attrName = tokens[1];
```
madf, line 72: How about `const std::string&`? Or even `const auto&`?

src/dictionaries.cpp
```diff
@@ -61,18 +69,29 @@ Dictionaries::Dictionaries(const std::string& filePath)
...
                 std::string code = tokens[2];
```
madf, line 73: How about calling `std::stoi` (or `std::stoul`) right here?

src/dictionaries.cpp
```diff
@@ -81,4 +100,23 @@ Dictionaries::Dictionaries(const std::string& filePath)
         }
         ++lineNumber;
     }
+//RFC
```
madf, line 103: What are these comments, from here and below?

madf (changes requested):

src/dictionaries.cpp
```diff
@@ -61,24 +62,44 @@ Dictionaries::Dictionaries(const std::string& filePath)
         {
             if (tokens[0] == "ATTRIBUTE")
             {
-                std::string name = tokens[1];
-                std::string code = tokens[2];
-                if (vendorFlag)
-                    std::cout << "  " << name << ": " << code << "\n";
+                const auto& attrName = tokens[1];
+                const auto& code = std::stoul(tokens[2]);
+                if (vendorName.size() != 0)
```
madf, line 67: Instead of checking string (or any other container) size against zero you better use `empty()` method. In some cases `size()` has linear complexity, when `empty()` is always guaranteed to have constant complexity. Your condition would be:
```c++
if (!vendorName.empty())
```
Also, it is semantically more correct.

madf (changes requested):

madf (changes requested):

PR #13, opened 2023-04-25
Simple RADIUS server

HelenMamontova:
Implemented processing of the $INCLUDE directive and filling all dictionaries.

### DIFF
--- a/src/dictionaries.cpp
+++ b/src/dictionaries.cpp
@@ -100,7 +100,14 @@ Dictionaries::Dictionaries(const std::string& filePath)
             else if (tokens[0] == "$INCLUDE")
             {
                 if (tokens[1].substr(0, 1) == "/")
+                {
                     Dictionaries fillingDictionaries(tokens[1]);
+                    m_attributes.append(fillingDictionaries.m_attributes);
+                    m_vendorNames.append(fillingDictionaries.m_vendorNames);
+                    m_attributeValues.append(fillingDictionaries.m_attributeValues);
+                    m_vendorAttributes.append(fillingDictionaries.m_vendorAttributes);
+                    m_vendorAttributeValues.append(fillingDictionaries.m_vendorAttributeValues);
+                }
                 else
                 {
                     Dictionaries fillingDictionaries(filePath.substr(0, filePath.rfind('/') + 1) + tokens[1]);
--- a/src/dictionaries.cpp
+++ b/src/dictionaries.cpp
@@ -15,16 +15,16 @@ uint32_t BasicDictionary::code(const std::string& name) const
     return m_reverseDict.at(name);
 }
 
-void BasicDictionary::append(const BasicDictionary& basicDict)
-{
-    m_rightDict.insert(basicDict.m_rightDict.begin(), basicDict.m_rightDict.end());
-    m_reverseDict.insert(basicDict.m_reverseDict.begin(), basicDict.m_reverseDict.end());
-}
-
 void BasicDictionary::add(uint32_t code, const std::string& name)
 {
     m_rightDict.emplace(code, name);
     m_reverseDict.emplace(name, code);
+}
+
+void BasicDictionary::append(const BasicDictionary& basicDict)
+{
+    m_rightDict.insert(basicDict.m_rightDict.begin(), basicDict.m_rightDict.end());
+    m_reverseDict.insert(basicDict.m_reverseDict.begin(), basicDict.m_reverseDict.end());
 }
 
 std::string DependentDictionary::name(const std::string& dependencyName, uint32_t code) const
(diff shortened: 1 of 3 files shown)


src/dictionaries.cpp
```diff
@@ -85,6 +85,13 @@ Dictionaries::Dictionaries(const std::string& filePath)
                 vendorName = tokens[1];
             else if (tokens[0] == "END-VENDOR")
                 vendorName.clear();
+            else if (tokens[0] == "$INCLUDE")
+            {
+                if (tokens[1].substr(0, 1) == "/")
+                    Dictionaries fillingDictionaries(tokens[1]);
+                else
+                    Dictionaries fillingDictionaries(filePath.substr(0, filePath.rfind('/') + 1)+ tokens[1]);
```
madf, line 93: Okay, you parsed the file in a local dictionary. Now what? Everything you parsed there is lost, the variable gets destroyed when it leaves the scope.

What we need here is to take all that data and put it into our local dictionaries.

madf (changes requested):

src/dictionaries.cpp
```diff
@@ -21,6 +21,12 @@ void BasicDictionary::add(uint32_t code, const std::string& name)
     m_reverseDict.emplace(name, code);
 }
 
+void BasicDictionary::append(const BasicDictionary& basicDict)
+{
+    m_rightDict.insert(basicDict.m_rightDict.begin(), basicDict.m_rightDict.end());
+    m_reverseDict.insert(basicDict.m_reverseDict.begin(), basicDict.m_reverseDict.end());
```
madf, line 27: The idea is correct, but there is a problem with this solution. If we are inserting a value that already exists in our dictionary, it won't be inserted or assigned. E.g., if we have a mapping
```
Some-Attribute 5
```
and then in the included dictionary we have
```
Some-Attribute 25
```
in the result of the `append` call we would end up with
```
Some-Attribute 5
```
So it does not change.

But we need to be able to override mappings, that the latter value has a precedence over the former.

src/dictionaries.cpp
```diff
@@ -85,21 +97,21 @@ Dictionaries::Dictionaries(const std::string& filePath)
                 vendorName = tokens[1];
             else if (tokens[0] == "END-VENDOR")
                 vendorName.clear();
+            else if (tokens[0] == "$INCLUDE")
+            {
+                if (tokens[1].substr(0, 1) == "/")
+                    Dictionaries fillingDictionaries(tokens[1]);
```
madf, line 103: Where are `append` calls for this?

madf (changes requested):

src/dictionaries.cpp
```diff
@@ -21,6 +21,15 @@ void BasicDictionary::add(uint32_t code, const std::string& name)
     m_reverseDict.emplace(name, code);
 }
 
+void BasicDictionary::append(const BasicDictionary& basicDict)
+{
+    for (const auto &entry: basicDict.m_rightDict)
```
madf, line 26: Must be `const auto& entry:`. Same below.

src/dictionaries.cpp
```diff
@@ -85,21 +103,28 @@ Dictionaries::Dictionaries(const std::string& filePath)
...
+                    m_vendorAttributeValues.append(fillingDictionaries.m_vendorAttributeValues);
+                }
+                else
+                {
+                    Dictionaries fillingDictionaries(filePath.substr(0, filePath.rfind('/') + 1) + tokens[1]);
+                    m_attributes.append(fillingDictionaries.m_attributes);
+                    m_vendorNames.append(fillingDictionaries.m_vendorNames);
+                    m_attributeValues.append(fillingDictionaries.m_attributeValues);
+                    m_vendorAttributes.append(fillingDictionaries.m_vendorAttributes);
+                    m_vendorAttributeValues.append(fillingDictionaries.m_vendorAttributeValues);
```
madf, line 124: I think it makes sense to add `append` method to the `Dictionaries` class. It will contain all these `append` calls for its members. And after adding the code above can be written as:
```c++
if (tokens[1].substr(0, 1) == "/")
    append(Dictionaries(tokens[1]));
else
    append(Dictionaries(filePath.substr(0, filePath.rfind('/') + 1) + tokens[1]));
```

madf (changes requested):

src/server.cpp
```diff
@@ -46,9 +46,26 @@ void printPacket(const Packet& p)
...
       : m_socket(io_service, udp::endpoint(udp::v4(), 9999)),
         m_secret(secret),
-        m_dictionaries("/usr/share/freeradius/dictionary.cisco")
-//        m_dictionaries("/usr/share/freeradius/dictionary.rfc2865")
+        m_dictionaries("/usr/share/freeradius/dictionary")
 {
+/*    for (const auto &entry: m_dictionaries.attributes().reverseDict())
+        std::cout << std::to_string(entry.second) << ": " << entry.first << "\n";*/
+
+    for (const auto &entry: m_dictionaries.attributes().rightDict())
```
madf, line 54: Must be `const auto& entry`. Same below.

madf (changes requested):

PR #14, opened 2023-05-18
Simple RADIUS server.

HelenMamontova:
Implemented the use of dictionaries of class Dictionaries by the Radius server.


src/packet.cpp
```diff
@@ -28,7 +27,7 @@ Packet::Packet(const std::array<uint8_t, 4096>& m_recvBuffer, size_t bytes, cons
         const uint8_t attributeType = m_recvBuffer[attributeIndex];
         const uint8_t attributeLength = m_recvBuffer[attributeIndex + 1];
 
-        if (attributeType == VENDOR_SPECIFIC)
+        if (attributeType == dict.attributes().code("Vendor-Specific"))
```
madf, line 30: I think we can keep the original version, with check against VENDOR_SPECIFIC from the enumeration. No need to pull in the Dictionaries.

src/packet.cpp
```diff
@@ -40,10 +39,10 @@ Packet::Packet(const std::array<uint8_t, 4096>& m_recvBuffer, size_t bytes, cons
     bool messageAuthenticator = false;
     for (size_t i = 0; i < m_attributes.size(); ++i)
     {
-        if (m_attributes[i]->type() == EAP_MESSAGE)
+        if (m_attributes[i]->type() == dict.attributes().code("EAP-Message"))
```
madf, line 42: Same as above, we can keed checking with EAP_MESSAGE (and MESSAGE_AUTHENTICATOR below).

src/vendor_attribute.cpp
```diff
@@ -58,6 +58,5 @@ std::string VendorSpecific::toString() const
 
 std::string VendorSpecific::name() const
```
madf, line 59: Do we still need this method?

madf (changes requested):

PR #23, opened 2023-09-06
Async RADIUS server.

HelenMamontova:
Async-radius server code split into radproto library and server executable.


src/CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
 configure_file ( version.h.in version.h ESCAPE_QUOTES @ONLY )
 
-target_sources ( ${PROJECT_NAME} PRIVATE main.cpp server.cpp packet.cpp attribute.cpp vendor_attribute.cpp utils.cpp dictionaries.cpp)
+add_library ( radproto STATIC )
+
+target_sources ( radproto PRIVATE server.cpp packet.cpp attribute.cpp vendor_attribute.cpp utils.cpp dictionaries.cpp)
+
+target_link_libraries ( ${PROJECT_NAME} radproto )
+
+target_include_directories ( radproto PUBLIC ${CMAKE_CURRENT_SOURCE_DIR} )
```
madf, line 9: This statement is not needed. By default, you can include headers from the same directory without explicitly stating that to a compiler. This statement would be useful if you use IMPORTED targets. For now you only need it for the sample server target.

src/CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
...
 
 target_link_libraries ( ${PROJECT_NAME} OpenSSL::Crypto Boost::boost Threads::Threads )
```
madf, line 11: This must go to the sample server.

madf: However you also need these imported targets in `radproto` library. Linkage does not happen here, but imported targets also provide include directories and compiler flags (`Threads::Threads' provide `-pthread` flag). Compilation may fail if OpenSSL or Boost are installed in non-standard places.

madf (changes requested):

src/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 configure_file ( version.h.in version.h ESCAPE_QUOTES @ONLY )
 
-target_sources ( ${PROJECT_NAME} PRIVATE main.cpp server.cpp packet.cpp attribute.cpp vendor_attribute.cpp utils.cpp dictionaries.cpp)
+add_library ( radproto STATIC )
+
+target_sources ( radproto PRIVATE server.cpp packet.cpp attribute.cpp vendor_attribute.cpp utils.cpp dictionaries.cpp)
+
+target_include_directories ( ${PROJECT_NAME} PUBLIC ${CMAKE_CURRENT_SOURCE_DIR} )
```
madf, line 7: This statement should go to the sample server CMakeLists.txt.

src/CMakeLists.txt
madf: There should be no mentions of the sample server in this CMakeLists.txt (`${PROJECT_NAME}`). This statement should relate to the library.

madf (changes requested):

sample/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+add_executable ( ${PROJECT_NAME} main.cpp )
+
+target_link_libraries ( ${PROJECT_NAME} radproto )
+
+target_link_libraries ( ${PROJECT_NAME} OpenSSL::Crypto Boost::boost Threads::Threads )
+
+target_include_directories ( ${PROJECT_NAME} PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../src )
```
madf, line 7: It is better not to deal with relative paths and use `CMAKE_SOURCE_DIR`, which points to the project root.

madf (changes requested):

PR #24, opened 2023-09-15
Async RADIUS server.

HelenMamontova:
Class header files moved to the async-radius/include/radproto directory.


src/attribute.cpp
```diff
@@ -1,5 +1,5 @@
-#include "packet.h"
-#include "attribute.h"
+#include "../include/radproto/packet.h"
```
madf, line 1: Since you added `target_include_directories`, relative paths are unnecessary here and below.

CMakeLists.txt
```diff
@@ -23,3 +23,8 @@ add_subdirectory (sample)
 add_subdirectory (src)
 
 add_custom_target (cppcheck COMMAND cppcheck --enable=all --std=c++14 ${CMAKE_SOURCE_DIR}/src)
+
+target_include_directories (radproto INTERFACE
```
madf, line 27: I think this directive must be in `src/CMakeLists.txt`.

madf (changes requested):

PR #29, opened 2023-11-15
Async RADIUS server

HelenMamontova:
The methods asyncReceive and asyncSend with callback added.


include/radproto/server.h
```diff
@@ -4,20 +4,20 @@
...
-#include <array>
 #include <cstdint> //uint8_t, uint32_t
+#include <array>
+#include <functional>
 
 class Server
 {
     public:
         Server(boost::asio::io_service& io_service, const std::string& secret);
+        void asyncReceive(std::function<void(const boost::system::error_code& error, const Packet& request)> callback);
```
madf, line 15: `callback` should be passed by const reference.

madf: Also, you do not need to name callback arguments, just types are enough.

include/radproto/server.h
```diff
@@ -4,20 +4,20 @@
...
+        void asyncSend(std::function<void(const boost::system::error_code& err)> callback, Packet& response);
```
madf, line 16: `response` must be a const reference.

src/server.cpp
```diff
@@ -24,95 +22,38 @@ std::string packetTypeToString(int type)
...
-    }
-}
-
 Server::Server(boost::asio::io_service& io_service, const std::string& secret)
-      : m_socket(io_service, udp::endpoint(udp::v4(), 9999)),
-        m_secret(secret),
-        m_dictionaries("/usr/share/freeradius/dictionary")
+    : m_socket(io_service, udp::endpoint(udp::v4(), 9999)),
+      m_secret(secret),
+      m_dictionaries("/usr/share/freeradius/dictionary")
```
madf, line 28: I guess, we no longer need dictionaries here.

include/radproto/server.h
madf: Also, let's make `response` the first argument.

src/server.cpp
```diff
@@ -24,95 +22,38 @@ std::string packetTypeToString(int type)
...
     try
     {
-        Packet packet = makeResponse(Packet(m_recvBuffer, bytes, m_secret));
-
-        std::cout << "Response packet\n";
-        printPacket(packet);
-
-        m_socket.async_send_to(boost::asio::buffer(packet.makeSendBuffer(m_secret)), m_remoteEndpoint,
-           [this](const error_code& /*ec*/, std::size_t /*bytesTransferred*/) {handleSend();});
+        const Packet request = Packet(m_recvBuffer, bytes, m_secret);
```
madf, line 48: You used to check error code and number of bytes before. You can not create a packet if the operation resulted in error, or if we received less than 20 bytes.

madf (changes requested):

src/error.cpp
```diff
@@ -0,0 +1,38 @@
+#include "error.h"
+#include <string>
+
+namespace lib
```
madf, line 4: Let's call the namespace `RadProto` and put everything in it, including `Packet`, `Server`, attributes and dictionaries.

src/error.cpp
```diff
@@ -0,0 +1,38 @@
...
+{
+    const char* ErrorCategory::name() const noexcept
+    {
+        return "library";
+    }
+
+    std::string ErrorCategory::message(int ev) const noexcept
+    {
+        switch(static_cast<Error>(ev))
```
madf, line 13: Missing whitespace after `switch`.

src/error.cpp
```diff
@@ -0,0 +1,38 @@
...
+            case Error::requestLengthShort:
+                return "Request length less than specified in the request";
+            case Error::eapMessageAttributeError:
+                return "The EAP-Message attribute is present, but the Message-Authenticator attribute is missing";
+            default:
+                return "(Unrecognized error)";
+        }
+    }
+
+    boost::system::error_category const& theCategory()
```
madf, line 28: We usually put `const` in front of the type.

src/error.h
```diff
@@ -0,0 +1,38 @@
+#ifndef ERROR_H
+#define ERROR_H
+
+#include <boost/system/error_code.hpp>
+
+namespace lib
```
madf, line 6: Same as above, `RadProto` namespace.

src/packet.cpp
```diff
@@ -7,13 +8,13 @@
 Packet::Packet(const std::array<uint8_t, 4096>& m_recvBuffer, size_t bytes, const std::string& secret)
 {
     if (bytes < 20)
-        throw std::runtime_error{"The number of received bytes in the request - " + std::to_string(bytes) + " is less than 20 bytes"};
+        throw std::runtime_error{"lib::Error::numberBytesLess20"};
```
madf, line 11: Instead of throwing `std::runtime_error`, please make a dedicated exception class, inherited from `std::runtime_error`. It must receive `Error` enum in the constructor, create an internal `error_code` member and use it to make a message for the `std::runtime_error` constructor.

madf: This line should look like this:
```c++
throw Exception(Error::numberBytesLessThan20);
```

src/packet.cpp
```diff
@@ -7,13 +8,13 @@
...
 
     size_t length = m_recvBuffer[2] * 256 + m_recvBuffer[3];
     std::cout << "Length: " << length << "\n";
 
     if (bytes < length)
-        throw std::runtime_error{"Request length " + std::to_string(bytes) + " is less than specified in the request - " + std::to_string(length)};
+        throw std::runtime_error{"requestLengthShort"};
```
madf, line 17: Same as above, use a dedicated class and `Error` enum.

src/packet.cpp
```diff
@@ -47,7 +48,7 @@ Packet::Packet(const std::array<uint8_t, 4096>& m_recvBuffer, size_t bytes, cons
             messageAuthenticator = true;
     }
     if (eapMessage && !messageAuthenticator)
-        throw std::runtime_error{"The EAP-Message attribute is present, but the Message-Authenticator attribute is missing"};
+        throw std::runtime_error{"eapMessageAttributeError"};
```
madf, line 51: Same as above.

src/server.cpp
```diff
@@ -40,23 +41,31 @@ void Server::asyncSend(const Packet& response, const std::function<void(const er
...
 }
 
-void Server::handleReceive(const error_code& error, std::size_t bytes, std::function<void(const error_code&, const Packet&)> callback)
+void Server::handleReceive(const error_code& error, std::size_t bytes, std::function<void(const error_code&, const std::optional<Packet>&)> callback)
 {
-    try
-    {
-        if (error)
-            throw std::runtime_error{"Error async_receivw_from: " + error.message()};
+        std::optional<Packet> oEmpty;
```
madf, line 46: You do not need `oEmpty`. Use `std::nullopt` instead.

src/server.cpp
```diff
@@ -40,23 +41,31 @@ void Server::asyncSend(const Packet& response, const std::function<void(const er
...
+        std::optional<Packet> oRequest;
```
madf, line 47: You do not need this variable, use `std::make_optional` where you pass the packet into the callback.

src/server.cpp
```diff
@@ -40,23 +41,31 @@ void Server::asyncSend(const Packet& response, const std::function<void(const er
...
+    try
+    {
         const Packet request = Packet(m_recvBuffer, bytes, m_secret);
-        callback(error, request);
+        oRequest = request;
     }
     catch (const std::runtime_error& exception)
     {
-        return;
+        if (std::string (exception.what()) == "lib::Error::numberBytesLess20")
```
madf, line 59: Never ever use string dispatch. Instead, as we discussed above, catch your dedicated exception class and extract `error_code` from it directly.

src/error.cpp
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    std::string ErrorCategory::message(int ev) const noexcept
+    {
+        switch(static_cast<Error>(ev))
+        {
+            case Error::success:
+                return "No error";
+            case Error::numberBytesLess20:
+                return "Request length less than 20 bytes";
```
madf, line 18: 'Request length is less than...' - missing 'is'. Same below.

src/error.h
```diff
@@ -0,0 +1,38 @@
...
+        public:
+            const char* name() const noexcept;
+            std::string message(int ev) const noexcept;
+    };
+
+    enum class Error
+    {
+        success = 0,
+
+        numberBytesLess20,
```
madf, line 19: `numberOfBytesIsLessThan20`

src/error.h
```diff
@@ -0,0 +1,38 @@
...
+
+    boost::system::error_code make_error_code(Error);
+}
+
+namespace boost
+{
+    namespace system
+    {
+        template <>
+            struct is_error_code_enum<::lib::Error> : std::true_type
```
madf, line 32: Wrong indentation. Put the `struct` keyword at the same distance as `template`.

madf (changes requested):

src/error.h
```diff
@@ -16,11 +16,21 @@ namespace lib
...
     };
 
+    class Exception: public std::runtime_error
+    {
+        public:
+            Exception(const boost::system::error_code& errorCode);
+            Error getErrorCode() const {return m_errorCode;}
+
+        private:
+            Error m_errorCode;
```
madf, line 31: Why do you use `Error` here if you get a `boost::system::error_code` and later convert the exception into the same `boost::system::error_code`?

src/server.cpp
```diff
@@ -2,73 +2,71 @@
...
+        if (bytes < 20)
+            callback(Error::numberOfBytesIsLessThan20, std::nullopt);
+        try
+        {
+            const Packet request = Packet(m_recvBuffer, bytes, m_secret);
+            callback(error, std::make_optional<Packet>(request));
+        }
+        catch (const Exception& exception)
+        {
+            if (exception.getErrorCode() == Error::numberOfBytesIsLessThan20)
```
madf, line 57: Why do you perform dispatch on error code? You can pass `exception.getErrorCode()` directly into the callback.

madf (changes requested):

src/error.cpp
```diff
@@ -0,0 +1,43 @@
+#include "error.h"
+#include <string>
+
+using Exception = RadProto::Exception;
+Exception::Exception(const boost::system::error_code& errorCode)
+    : std::runtime_error(errorCode.message())
```
madf, line 6: Missing `m_errorCode` initialization.

src/server.cpp
```diff
@@ -12,107 +11,53 @@ std::string packetTypeToString(int type)
...
     {
-        Packet packet = makeResponse(Packet(m_recvBuffer, bytes, m_secret));
-
-        std::cout << "Response packet\n";
-        printPacket(packet);
-
-        m_socket.async_send_to(boost::asio::buffer(packet.makeSendBuffer(m_secret)), m_remoteEndpoint,
-           [this](const error_code& /*ec*/, std::size_t /*bytesTransferred*/) {handleSend();});
+        const Packet request = Packet(m_recvBuffer, bytes, m_secret);
+        callback(error, std::make_optional<Packet>(request));
```
madf, line 52: You do not need to create `Packet` first and put it into `std::make_optional`. This function works as a constructor, it accepts a template parameter (e.g. `Packet`) and constructs it inside using arguments. Consider the following two examples:
```c++
const Packet request = Packet(buffer, bytes, secret); // Constructs 'request' object using 'buffer', 'bytes' and 'secret'
const auto op1 = std::make_optional<Packet>(); // Constructs an object of type 'Packet' using 'request' as an argumet (calls copy constructor)

// ---***---

// Here is a better way
const auto op2 = std::make_optional<Packet>(buffer, bytes, secret); // Constructs an object of type 'Packet' using 'buffer', 'bytes' and 'secret'
```

src/server.cpp
```diff
@@ -12,107 +11,53 @@ std::string packetTypeToString(int type)
...
-    if (error)
-    {
-        std::cout << "Error async_receivw_from: " << error.message() << "\n";
-        return;
-    }
+    m_socket.async_send_to(boost::asio::buffer(response.makeSendBuffer(m_secret)), m_remoteEndpoint,
+       [this, callback](const error_code& ec, std::size_t /*bytesTransferred*/) {handleSend(ec, callback);});
+}
 
+void Server::handleReceive(const error_code& error, std::size_t bytes, std::function<void(const error_code&, const std::optional<Packet>&)> callback)
```
madf, line 45: The `callback` should be passed by const reference

src/server.cpp
```diff
@@ -12,107 +11,53 @@ std::string packetTypeToString(int type)
...
     }
-    startReceive();
-}
-
-void Server::handleSend()
-{
 }
 
-Packet Server::makeResponse(const Packet& request)
+void Server::handleSend(const error_code& ec, std::function<void(const error_code&)> callback)
```
madf, line 60: The `callback` should be passed by const reference

src/error.cpp
```diff
@@ -0,0 +1,43 @@
...
+        case Error::requestLengthIsShort:
+            return "Request length is less than specified in the request";
+        case Error::eapMessageAttributeError:
+            return "The EAP-Message attribute is present, but the Message-Authenticator attribute is missing";
+        default:
+            return "(Unrecognized error)";
+    }
+}
+
+const boost::system::error_category& theCategory()
```
madf, line 34: Let's ditch `theCategory` function and put the `instance` directly inside `make_error_code`,

madf (changes requested):

src/error.h
```diff
@@ -0,0 +1,48 @@
...
+
+        numberOfBytesIsLessThan20,
+        requestLengthIsShort,
+        eapMessageAttributeError
+    };
+
+    class Exception: public std::runtime_error
+    {
+        public:
+            Exception(const boost::system::error_code& errorCode);
```
madf, line 27: Let's make this constructor explicit.

sample/main.cpp
```diff
@@ -64,7 +64,7 @@ int main(int argc, char* argv[])
     try
     {
         boost::asio::io_service io_service;
-        Server server(io_service, secret);
+        RadProto::Server server(io_service, secret);
```
madf, line 67: After the changes in the library, this sample no longer works. You need to update the sample code to make it work like before.

madf (changes requested):

src/radius.cpp
```diff
@@ -0,0 +1,65 @@
...
+        case RadProto::ACCESS_CHALLENGE: return "ACCESS_CHALLENGE";
+        case RadProto::STATUS_SERVER: return "STATUS_SERVER";
+        case RadProto::STATUS_CLIENT: return "STATUS_CLIENT";
+    }
+    return "uncnown";
+}
+
+using Radius = RadProto::Radius;
+Radius::Radius(boost::asio::io_service& io_service, const std::string& secret)
+    : m_socket(io_service, udp::endpoint(udp::v4(), 9999)),
```
madf, line 28: Port number must be configurable, with default value 1812.

madf (changes requested):

include/radproto/attribute.h
```diff
@@ -6,85 +6,95 @@
...
-        virtual std::vector<uint8_t> toVector(const std::string& secret, const std::array<uint8_t, 16>& auth) const = 0;
+    class Attribute
+    {
+        public:
+            Attribute(uint8_t type);
+            virtual ~Attribute() = default;
+            uint8_t type() const { return m_type; }
+            virtual std::string toString() const = 0;
+            virtual std::vector<uint8_t> toVector(const std::string& secret, const std::array<uint8_t, 16>& auth) const = 0;
+             virtual Attribute* clone() const = 0;
```
madf, line 19: Indentation problem.

include/radproto/packet.h
```diff
@@ -8,26 +8,30 @@
...
-        const std::vector<Attribute*>& attributes() const { return m_attributes; }
-        const std::vector<VendorSpecific*>& vendorSpecific() const { return m_vendorSpecific; }
-        const std::vector<uint8_t> makeSendBuffer(const std::string& secret);
-        Attribute* makeAttribute(uint8_t type, const uint8_t* data, size_t size, const std::string& secret, const std::array<uint8_t, 16>& auth);
+    class Packet
+    {
+        public:
+            Packet(const std::array<uint8_t, 4096>& m_recvBuffer, size_t bytes, const std::string& secret);
+            Packet(uint8_t type, uint8_t id, const std::array<uint8_t, 16>& auth, const std::vector<Attribute*>& attributes, const std::vector<VendorSpecific*>& vendorSpecific);
+            Packet(const Packet& dc);
```
madf, line 18: What does `dc` mean? If you have no better name, consider `other` or `rhs` (which stands for 'right-hand side').

include/radproto/packet.h
```diff
@@ -8,26 +8,30 @@
...
-        uint8_t id() const { return m_id; };
-        const std::array<uint8_t, 16>& auth() const { return m_auth; }
-        const std::vector<Attribute*>& attributes() const { return m_attributes; }
-        const std::vector<VendorSpecific*>& vendorSpecific() const { return m_vendorSpecific; }
-        const std::vector<uint8_t> makeSendBuffer(const std::string& secret);
-        Attribute* makeAttribute(uint8_t type, const uint8_t* data, size_t size, const std::string& secret, const std::array<uint8_t, 16>& auth);
+    class Packet
+    {
+        public:
+            Packet(const std::array<uint8_t, 4096>& m_recvBuffer, size_t bytes, const std::string& secret);
```
madf, line 16: You should not call function arguments `m_<something>`. This scheme is used for class members. Let it be just `buffer`.

sample/main.cpp
```diff
@@ -40,12 +47,21 @@ int main(int argc, char* argv[])
             printHelp(argv[0]);
             return 0;
         }
-        if (arg == "--version" || arg == "-v")
+        else if (arg == "--version" || arg == "-v")
```
madf, line 50: You do not need `else` if the previous `if` ends with `return`.

sample/main.cpp
```diff
@@ -40,12 +47,21 @@ int main(int argc, char* argv[])
...
         }
-        if (arg == "--secret" || arg == "-s")
+        else if (arg == "--port" || arg == "-p")
+        {
+            if (i + 1 == argc)
+            {
+                std::cerr << arg << " required argument - port number.\n";
+                return 1;
+            }
+            port = atoi(argv[++i]);
```
madf, line 62: `atoi` is awful, it does not check enything. Let's use `std::stoul` or `std::from_chars`.

src/packet.cpp
```diff
@@ -59,13 +59,27 @@ Packet::Packet(uint8_t type, uint8_t id, const std::array<uint8_t, 16>& auth, co
 {
 }
 
+Packet::Packet(const Packet& dc)
+{
+    for(size_t i = 0; i < dc.m_vendorSpecific.size(); ++i)
+        if (dc.m_vendorSpecific[i])
+            m_vendorSpecific.push_back(new VendorSpecific(*dc.m_vendorSpecific[i]));
```
madf, line 66: How about:
```c++
for (const a& : dc.m_vendorSpecific)
    if (a)
        m_vendorSpecific.push_back(new VendorSpecific(*a));
```

Same for `m_attributes` below.

src/radius.cpp
```diff
@@ -0,0 +1,65 @@
...
+    const std::vector<uint8_t> vResponse = response.makeSendBuffer(m_secret);
+    std::copy(vResponse.begin(), vResponse.end(), m_recvBuffer.begin());
+
+    m_socket.async_send_to(boost::asio::buffer(m_recvBuffer), m_remoteEndpoint,
+       [this, callback](const error_code& ec, std::size_t /*bytesTransferred*/) {handleSend(ec, callback);});
+}
+
+void Radius::handleReceive(const error_code& error, std::size_t bytes, const std::function<void(const error_code&, const std::optional<Packet>&)>& callback)
+{
+    if (bytes < 20)
```
madf, line 50: Let's check for `error` first.

madf (changes requested):

src/radius.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+    m_socket.async_send_to(boost::asio::buffer(m_recvBuffer), m_remoteEndpoint,
+       [this, callback](const error_code& ec, std::size_t /*bytesTransferred*/) {handleSend(ec, callback);});
+}
+
+void Radius::handleReceive(const error_code& error, std::size_t bytes, const std::function<void(const error_code&, const std::optional<Packet>&)>& callback)
+{
+    try
+    {
+        callback(error, std::make_optional<Packet>(m_recvBuffer, bytes, m_secret));
```
madf, line 52: What if the receive operation ended with error? You do not check for errors here, you try to make a packet right away, even if the `m_buffer` contains garbage.

src/radius.cpp
```diff
@@ -0,0 +1,65 @@
...
+    }
+    catch (const Exception& exception)
+    {
+        callback(exception.getErrorCode(), std::nullopt);
+    }
+    if (bytes < 20)
```
madf, line 58: This check is useless. It is also dangerous, because if you received less than 20 bytes, you call the `callback` twice, which is wrong.

include/radproto/packet.h
```diff
@@ -8,26 +8,30 @@
...
-        std::array<uint8_t, 16> m_auth;
-        std::vector<Attribute*> m_attributes;
-        std::vector<VendorSpecific*> m_vendorSpecific;
-};
+        private:
+            uint8_t m_type;
+            uint8_t m_id;
+            std::array<uint8_t, 16> m_auth;
+            std::vector<Attribute*> m_attributes;
+            std::vector<VendorSpecific*> m_vendorSpecific;
```
madf, line 33: Do we really need to store `VendorSpecific` by pointers? Can we store them simply by value?

madf (changes requested):

src/packet.cpp
```diff
@@ -59,13 +59,21 @@ Packet::Packet(uint8_t type, uint8_t id, const std::array<uint8_t, 16>& auth, co
 {
 }
 
+Packet::Packet(const Packet& other)
+    : m_vendorSpecific(other.m_vendorSpecific)
+{
+    for(const auto& a :  other.m_attributes)
```
madf, line 65: Missing whitespace after `for`.

src/packet.cpp
```diff
@@ -103,7 +111,7 @@ const std::vector<uint8_t> Packet::makeSendBuffer(const std::string& secret)
     return sendBuffer;
 }
 
-Attribute* Packet::makeAttribute(uint8_t type, const uint8_t* data, size_t size, const std::string& secret, const std::array<uint8_t, 16>& auth)
+RadProto::Attribute* Packet::makeAttribute(uint8_t type, const uint8_t* data, size_t size, const std::string& secret, const std::array<uint8_t, 16>& auth)
```
madf, line 114: This function does not use any class members, so it should be either a free function, or a static member. It is only used in `Packet` constructor, so, ideally, it should be a local function in the anonymous namespace.

madf (changes requested):

madf (changes requested):

include/radproto/vendor_attribute.h
```diff
@@ -5,18 +5,22 @@
...
-        std::string toString() const;
-        uint8_t vendorType() const { return m_vendorType; }
-        uint32_t vendorId() const { return m_vendorId; }
-        std::vector<uint8_t> toVector() const;
-    private:
-        uint32_t m_vendorId;
-        uint8_t m_vendorType;
-        std::vector<uint8_t> m_value;
-};
+    class VendorSpecific
```
madf, line 10: There is a problem with this class. Consider your sample server reply:
```
Sent Access-Request Id 117 from 0.0.0.0:60875 to 127.0.0.1:1812 length 74
        User-Name = "test"
        User-Password = "123456"
        NAS-IP-Address = 127.0.0.1
        NAS-Port = 1
        Message-Authenticator = 0x00
        Cleartext-Password = "123456"
Received Access-Accept Id 117 from 127.0.0.1:1812 to 127.0.0.1:60875 length 95
        User-Name = "test"
        NAS-Port = 20
        NAS-IP-Address = 127.104.22.17
        User-Password = "password123"
        Callback-Number = "123abc"
        CHAP-Password = 0x0131323334353637383961626364656667
        Dlink-User-Level = 808452096
```

You can observe that 'Dlink-User-Level' is rendered as an integer value (it is indeed integer according to the dictionary). You set it as the following vector of bytes:
```c++
std::vector<uint8_t> vendorValue {'0', '0', '0', '3'};
```
It looks strange because you use digits as characters rather than values (i.e., `{0, 0, 0, 3}`), but its okay. If we take these bytes and turn them into a big endian value, we get:
```
 - '0' has hex code 0x30 = 48;
 - '3' has hex code 0x33 = 51.
 
 48 * 256 * 256 * 256 +
 48 * 256 * 256 +
 48 * 256 +
 51 = 808464435 = 0x30303033
```
However, your sample server responds with 808452096 = 0x30300000

It seems that your `toVector` code takes only first two bytes from the value and skips the rest.

madf (changes requested):

include/radproto/socket.h
```diff
@@ -0,0 +1,31 @@
+#ifndef SOCKET_H
```
madf, line 1: `SOCKET_H` is rather common. Either add a library prefix (`RADPROTO_SOCKET_H`), or use more modern `#pragma once` instead of `#ifndef`/`#define`/`#endif`.

Make sure you do it in every header file.

madf (changes requested):

PR #33, opened 2024-04-11
Async Radius. Unit tests.

HelenMamontova:
Unit tests for class String added.

### DIFF
--- a/tests/attribute_tests.cpp
+++ b/tests/attribute_tests.cpp
@@ -1,6 +1,7 @@
 #define BOOST_TEST_MODULE RadProtoAttributeTests
 
 #include "radproto/attribute.h"
+#include <memory>
 
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wold-style-cast"
@@ -44,15 +45,16 @@ BOOST_AUTO_TEST_CASE(StringValueConstructor)
 BOOST_AUTO_TEST_CASE(StringClone)
 {
     RadProto::String c(1, "test");
+    std::unique_ptr<RadProto::Attribute> cs(c.clone());
 
-    BOOST_CHECK_EQUAL(c.clone()->toString(), "test");
+    BOOST_CHECK_EQUAL(cs->toString(), "test");
 
-    std::vector<uint8_t> values = c.clone()->toVector({}, {});
+    std::vector<uint8_t> values = cs->toVector({}, {});
     std::vector<uint8_t> expected {1, 6, 't', 'e', 's', 't'};
 
     BOOST_TEST(values == expected, boost::test_tools::per_element());
 
-    BOOST_CHECK_EQUAL(c.clone()->type(), 1);
+    BOOST_CHECK_EQUAL(cs->type(), 1);
 }
 
 BOOST_AUTO_TEST_SUITE_END()
--- a/tests/attribute_tests.cpp
+++ b/tests/attribute_tests.cpp
@@ -10,7 +10,7 @@
 #include <boost/test/unit_test.hpp>
 #pragma GCC diagnostic pop
 
-BOOST_AUTO_TEST_SUITE()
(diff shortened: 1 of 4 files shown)


madf (changes requested):

tests/attribute_tests.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wold-style-cast"
+#pragma GCC diagnostic ignored "-Wunused-parameter"
+#pragma GCC diagnostic ignored "-Wsign-compare"
+#pragma GCC diagnostic ignored "-Wparentheses"
+#include <boost/test/unit_test.hpp>
+#pragma GCC diagnostic pop
+
+BOOST_AUTO_TEST_SUITE(RadProtoAttributeTests)
```
madf, line 13: Just 'AttributeTests' would be fine.

tests/attribute_tests.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+BOOST_AUTO_TEST_CASE(ClassStringDataConstructor)
```
madf, line 15: 'StringDataConstructor' would be ok.

tests/attribute_tests.cpp
```diff
@@ -0,0 +1,58 @@
...
+{
+    std::vector<uint8_t> d {'t', 'e', 's', 't'};
+    RadProto::String s(1, d.data(), d.size());
+
+    BOOST_CHECK_EQUAL(s.toString(), "test");
+
+    std::vector<uint8_t> values = s.toVector({}, {});
+    std::vector<uint8_t> expected {1, 6, 't', 'e', 's', 't'};
+
+    BOOST_CHECK_EQUAL_COLLECTIONS(values.begin(), values.end(), expected.begin(), expected.end());
```
madf, line 25: Nice finding! But it is a little bit too old-fashioned. Let's use modern alternative:
```
BOOST_TEST(values == expected, boost::test_tools::per_element());
```
Details: https://www.boost.org/doc/libs/1_85_0/libs/test/doc/html/boost_test/utf_reference/testing_tool_ref/assertion_boost_test_universal_macro.html

tests/attribute_tests.cpp
madf: Same below.

tests/attribute_tests.cpp
madf: Same below.

tests/attribute_tests.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+    BOOST_CHECK_EQUAL(v.type(), 1);
+}
+
+BOOST_AUTO_TEST_CASE(ClassStringClone)
+{
+    RadProto::String c(1, "test");
+
+    BOOST_CHECK_EQUAL(c.clone()->toString(), "test");
```
madf, line 48: Although it is not a big problem here, remember that `clone()` makes a copy in heap. And you need to delete it manually.

1. Do not call `clone()` in test statements, make a single clone and use it.
2. Take care of the resource, delete it after use.

I propose something like this:
```c++
RadProto::String s(1, "test");
std::unique_ptr<RadProto::Attribute> cs(c.clone());

BOOST_CHECK_EQUAL(cs->toString(), "test");
// ...
```

tests/error_tests.cpp
```diff
@@ -10,13 +10,13 @@
 #include <boost/test/unit_test.hpp>
 #pragma GCC diagnostic pop
 
-BOOST_AUTO_TEST_SUITE()
+BOOST_AUTO_TEST_SUITE(RadProtoErrorTests)
```
madf, line 13: `ErrorTests` would be just fine.

madf (changes requested):

madf:
Looks good!

PR #35, opened 2024-04-19
Async Radius. Unit tests.

HelenMamontova:
IntegerDataConstructor, IntegerValueConstructor, IntegerClone test cases added.

### DIFF
--- a/tests/attribute_tests.cpp
+++ b/tests/attribute_tests.cpp
@@ -57,4 +57,48 @@ BOOST_AUTO_TEST_CASE(StringClone)
     BOOST_CHECK_EQUAL(cs->type(), 1);
 }
 
+BOOST_AUTO_TEST_CASE(IntegerDataConstructor)
+{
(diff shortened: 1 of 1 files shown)


tests/attribute_tests.cpp
```diff
@@ -57,4 +57,48 @@ BOOST_AUTO_TEST_CASE(StringClone)
     BOOST_CHECK_EQUAL(cs->type(), 1);
 }
 
+BOOST_AUTO_TEST_CASE(IntegerDataConstructor)
+{
+    std::vector<uint8_t> d {0, 0, 0, 20};
```
madf, line 62: This test value does not cover multibyte conversions. Let's use something more interesting. For example, `{10, 20, 30, 40}`.

madf: Same below.

madf (changes requested):

PR #37, opened 2024-05-03
Async Radius. Unit tests.

HelenMamontova:
EncryptedDataConstructor, EncryptedValueConstructor, EncryptedClone test cases added.


src/attribute.cpp
```diff
@@ -147,6 +148,9 @@ Encrypted::Encrypted(uint8_t type, const uint8_t* data, size_t size, const std::
             mdBuffer[j + secret.length()] = data[i * 16 + j];
     }
 
+    const auto deleteObject = std::remove(plaintext.begin(), plaintext.end(), 0);
+    plaintext.erase(deleteObject, plaintext.end());
```
madf, line 152: No need in `deleteObject` variable. Use it inline:
```c++
plaintext.erase(std::remove(plaintext.begin(), plaintext.end(), 0), plaintext.end());
```

This is classic C++ 'Remove-Erase' idiom.

madf: On the other hand, I think current behavior is still wrong. Consider a string like this: `"test\0abc"`. Your algorithm will produce `"testabc"`, when the correct should be just `"test"`.

Consider finding a first zero element in the range and use it instead of `plaintext.end()` in the `assign` below.

tests/attribute_tests.cpp
```diff
@@ -151,4 +151,51 @@ BOOST_AUTO_TEST_CASE(IpAddressClone)
     BOOST_CHECK_EQUAL(cs->type(), 4);
 }
 
+BOOST_AUTO_TEST_CASE(EncryptedDataConstructor)
+{
+    std::array<uint8_t, 16> auth {0x92, 0xfa, 0xa1, 0xed, 0x98, 0x9b, 0xb4, 0x79, 0xfe, 0x20, 0xe2, 0xf4, 0x7f, 0x4a, 0x5a, 0x70};
+    std::vector<uint8_t> d {0x25, 0x38, 0x58, 0x18, 0xae, 0x97, 0xeb, 0xeb, 0xbd, 0x46, 0xfd, 0xb9, 0xd1, 0x17, 0x84, 0xeb};
+    RadProto::Encrypted s(2, d.data(), d.size(), "secret", auth);
+
+    BOOST_CHECK_EQUAL(s.toString(), "123456");
```
madf, line 160: Let's also test passwords (`RadProto::Encrypted`) with 15-symbol length, 16-symbol length and 17-symbol length. Edge cases are important.

Also, how about empty passwords?

tests/attribute_tests.cpp
```diff
@@ -151,4 +151,51 @@ BOOST_AUTO_TEST_CASE(IpAddressClone)
...
+
+    std::vector<uint8_t> values = s.toVector("secret", auth);
+    std::vector<uint8_t> expected {0x02, 0x12, 0x25, 0x38, 0x58, 0x18, 0xae, 0x97, 0xeb, 0xeb, 0xbd, 0x46, 0xfd, 0xb9, 0xd1, 0x17, 0x84, 0xeb};
```
madf, line 163: `expected` should contain all `d` elements prefixed with `0x02, <length>`. How about making a copy of `d` in elements and using `push_front`? Or reserving a full length, putting `0x02, <length>` inside, and then using `std::copy`?

madf (changes requested):

src/attribute.cpp
```diff
@@ -147,7 +148,12 @@ Encrypted::Encrypted(uint8_t type, const uint8_t* data, size_t size, const std::
             mdBuffer[j + secret.length()] = data[i * 16 + j];
     }
 
-    m_value.assign(plaintext.begin(), plaintext.end());
+    auto it = std::find(plaintext.begin(), plaintext.end(), 0);
+
+    if (it != plaintext.end())
+        m_value.assign(plaintext.begin(), it);
+    else
+        m_value.assign(plaintext.begin(), plaintext.end());
```
madf, line 156: Why not just:
```c++
m_value.assign(plaintext.begin(), std::find(plaintext.begin(), plaintext.end(), 0));
```

madf (changes requested):

PR #39, opened 2024-05-21
Async Radius. Unit tests.

HelenMamontova:
ChapPasswordDataConstructor, ChapPasswordValueConstructor, ChapPasswordClone test cases added.

### DIFF
--- a/src/attribute.cpp
+++ b/src/attribute.cpp
@@ -261,7 +261,7 @@ std::string ChapPassword::toString() const
     for (const auto& b : m_value)
         value += byteToHex(b);
 
-    return std::to_string(m_chapId) + " " + value;
+    return byteToHex(m_chapId) + " " + value;
 }
 
 std::vector<uint8_t> ChapPassword::toVector(const std::string& /*secret*/, const std::array<uint8_t, 16>& /*auth*/) const
--- a/src/attribute.cpp
+++ b/src/attribute.cpp
@@ -261,7 +261,7 @@ std::string ChapPassword::toString() const
     for (const auto& b : m_value)
         value += byteToHex(b);
 
-    return byteToHex(m_chapId) + " " + value;
+    return std::to_string(m_chapId) + " " + value;
 }
 
 std::vector<uint8_t> ChapPassword::toVector(const std::string& /*secret*/, const std::array<uint8_t, 16>& /*auth*/) const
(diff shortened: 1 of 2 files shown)


src/attribute.cpp
```diff
@@ -261,7 +261,7 @@ std::string ChapPassword::toString() const
     for (const auto& b : m_value)
         value += byteToHex(b);
 
-    return std::to_string(m_chapId) + " " + value;
+    return byteToHex(m_chapId) + " " + value;
```
madf, line 264: Why do you change this method?

tests/attribute_tests.cpp
```diff
@@ -401,4 +401,48 @@ BOOST_AUTO_TEST_CASE(BytesClone)
     BOOST_CHECK_EQUAL(cs->type(), 19);
 }
 
+BOOST_AUTO_TEST_CASE(ChapPasswordDataConstructor)
+{
+    std::vector<uint8_t> d {'1', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f', 'g'};
```
madf, line 406: Why do you use chars here and hex codes below? It would be clearer to use either chars or hexes in both cases.

tests/attribute_tests.cpp
```diff
@@ -401,4 +401,48 @@ BOOST_AUTO_TEST_CASE(BytesClone)
...
+    RadProto::ChapPassword s(3, d.data(), d.size());
+
+    BOOST_CHECK_EQUAL(s.toString(), "31 31323334353637383961626364656667");
+
+    std::vector<uint8_t> values = s.toVector({}, {});
+    std::vector<uint8_t> expected({0x03, 0x13, 0x31, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x38, 0x39, 0x61, 0x62, 0x63, 0x64, 0x65, 0x66, 0x67});
```
madf, line 412: This vector almost identical to the `d` vector. Consider making a copy and prepending it with type and length, like you do it for other attribute test.

tests/attribute_tests.cpp
```diff
@@ -401,4 +401,48 @@ BOOST_AUTO_TEST_CASE(BytesClone)
...
+
+    BOOST_TEST(values == expected, boost::test_tools::per_element());
+
+    BOOST_CHECK_EQUAL(s.type(), 3);
+}
+
+BOOST_AUTO_TEST_CASE(ChapPasswordValueConstructor)
+{
+    RadProto::ChapPassword v(3, '1', {'1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f', 'g'});
```
madf, line 421: Why `chapId` is a char?

madf (changes requested):

PR #40, opened 2024-05-24
Async Radius. Unit tests.

HelenMamontova:
The test cases for to check throwing exeptions added for IntegerDataСonstructor, IpAddressDataConstructor, EncryptedDataConstructor,  ChapPasswordDataConstructor.


tests/attribute_tests.cpp
```diff
@@ -80,6 +80,13 @@ BOOST_AUTO_TEST_CASE(IntegerDataConstructor)
     BOOST_CHECK_EQUAL(s.type(), 5);
 }
 
+BOOST_AUTO_TEST_CASE(IntegerDataConstructorThrow)
+{
+    std::vector<uint8_t> d {10, 20, 30, 40, 60};
+
+    BOOST_CHECK_THROW(RadProto::Integer(5, d.data(), d.size()), std::runtime_error);
```
madf, line 87: That's interesting! You have `RadProto::Exception` and you use it in `Packet` and `Socket`, but here you really throw `std::runtime_error`. I think this is wrong, if we have library-level exception type, we should use it, rather than generic `std::runtime_error`. Consider throwing `Exception` here and below.

madf (changes requested):

PR #44, opened 2024-06-17
Async Radius. Unit test.

HelenMamontova:
The test case PacketBufferConstructor added.


tests/packet_tests.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+    std::array<uint8_t, 16> md;
+    MD5(withSecret.data(), withSecret.size(), md.data());
+
+    std::vector<uint8_t> expected({0x01, 0x7c, 0x00, 0x5c, 0x5d, 0xc1, 0x95, 0x4e, 0x15, 0xc8, 0xfd, 0x76, 0x41, 0x77, 0xbb, 0xfc, 0x6e, 0x5f, 0xda, 0x8a, 0x01, 0x06, 0x74, 0x65, 0x73, 0x74, 0x02, 0x12, 0xc1, 0x07, 0xf2, 0xcc, 0x60, 0x4f, 0x36, 0x03, 0x26, 0xe2, 0x05, 0x3d, 0x79, 0x5b, 0xe5, 0xd6, 0x04, 0x06, 0x7f, 0x00, 0x00, 0x01, 0x05, 0x06, 0x00, 0x00, 0x00, 0x01, 0x50, 0x12, 0xc3, 0xbe, 0x0e, 0x1a, 0xab, 0xcf, 0x2e, 0x7c, 0x7c, 0x00, 0x8c, 0x85, 0x9b, 0x32, 0x65, 0x3b, 0x07, 0x06, 0x00, 0x00, 0x00, 0x01, 0x1a, 0x0c, 0x00, 0x00, 0x00, 0xab, 0x01, 0x06, 0x00, 0x00, 0x00, 0x03});
+
+    for (size_t i = 0; i < md.size(); ++i)
+        expected[i + 4] = md[i];
+
+    BOOST_TEST(values == expected, boost::test_tools::per_element());
```
madf, line 122: Shouldn't the `values` be exactly the same as `d` - the data you used for construction?

madf (changes requested):

src/packet.cpp
```diff
@@ -79,11 +80,13 @@ Packet::Packet(uint8_t type, uint8_t id, const std::array<uint8_t, 16>& auth, co
       m_attributes(attributes),
       m_vendorSpecific(vendorSpecific)
 {
+    m_recalcAuth = true;
 }
 
 Packet::Packet(const Packet& other)
     : m_vendorSpecific(other.m_vendorSpecific)
 {
+    m_recalcAuth = false;
```
madf, line 89: Put it into the member initialization list.

madf: By the way, why is it `false`? Shouldn't it be the same as in `other`?

tests/packet_tests.cpp
```diff
@@ -105,19 +103,87 @@ BOOST_AUTO_TEST_CASE(PacketBufferConstructor)
     BOOST_CHECK(types.count(attrs[4]->type()) == 1);
     BOOST_CHECK(types.count(attrs[5]->type()) == 1);
 
+    std::vector<uint8_t> values = p.makeSendBuffer("secret");
+
+    BOOST_REQUIRE_EQUAL(values.size(), 92);
+
+    std::vector<uint8_t> expected({0x01, 0xd0, 0x00, 0x5c, 0x1a, 0x40, 0x43, 0xc6, 0x41, 0x0a, 0x08, 0x31, 0x12, 0x16, 0x80, 0x2c, 0x3e, 0x83, 0x12, 0x45, 0x01, 0x06, 0x74, 0x65, 0x73, 0x74, 0x02, 0x12, 0x8c, 0x06, 0xc8, 0x23, 0x55, 0xba, 0x0d, 0xd6, 0x15, 0x1c, 0xbf, 0x9d, 0xd8, 0x1a, 0x4d, 0x87, 0x04, 0x06, 0x7f, 0x00, 0x00, 0x01, 0x05, 0x06, 0x00, 0x00, 0x00, 0x01, 0x50, 0x12, 0xf3, 0xe0, 0x00, 0xe7, 0x7d, 0xeb, 0x51, 0xeb, 0x81, 0x5d, 0x52, 0x37, 0x3d, 0x06, 0xb7, 0x1b, 0x07, 0x06, 0x00, 0x00, 0x00, 0x01, 0x1a, 0x0c, 0x00, 0x00, 0x00, 0xab, 0x01, 0x06, 0x00, 0x00, 0x00, 0x03});
```
madf, line 110: I think this is the same as `d` and you should use the `d` in the comparison below.

madf (changes requested):

tests/packet_tests.cpp
```diff
@@ -107,7 +108,10 @@ BOOST_AUTO_TEST_CASE(PacketBufferConstructorRequest)
 
     BOOST_REQUIRE_EQUAL(values.size(), 92);
 
-    std::vector<uint8_t> expected({0x01, 0xd0, 0x00, 0x5c, 0x1a, 0x40, 0x43, 0xc6, 0x41, 0x0a, 0x08, 0x31, 0x12, 0x16, 0x80, 0x2c, 0x3e, 0x83, 0x12, 0x45, 0x01, 0x06, 0x74, 0x65, 0x73, 0x74, 0x02, 0x12, 0x8c, 0x06, 0xc8, 0x23, 0x55, 0xba, 0x0d, 0xd6, 0x15, 0x1c, 0xbf, 0x9d, 0xd8, 0x1a, 0x4d, 0x87, 0x04, 0x06, 0x7f, 0x00, 0x00, 0x01, 0x05, 0x06, 0x00, 0x00, 0x00, 0x01, 0x50, 0x12, 0xf3, 0xe0, 0x00, 0xe7, 0x7d, 0xeb, 0x51, 0xeb, 0x81, 0x5d, 0x52, 0x37, 0x3d, 0x06, 0xb7, 0x1b, 0x07, 0x06, 0x00, 0x00, 0x00, 0x01, 0x1a, 0x0c, 0x00, 0x00, 0x00, 0xab, 0x01, 0x06, 0x00, 0x00, 0x00, 0x03});
+    std::vector<uint8_t> expected;
```
madf, line 111: You do not need a cycle to create a copy of an array as a vector.

```c++
std::vector<uint8_t> expected(d.begin(), d.end());
```

madf: Same below.

src/packet.cpp
```diff
@@ -75,14 +76,17 @@ Packet::Packet(const std::array<uint8_t, 4096>& buffer, size_t bytes, const std:
...
+      m_recalcAuth(true),
       m_auth(auth),
       m_attributes(attributes),
       m_vendorSpecific(vendorSpecific)
 {
 }
 
 Packet::Packet(const Packet& other)
-    : m_vendorSpecific(other.m_vendorSpecific)
+    : m_recalcAuth(true),
```
madf, line 87: Why `m_recalcAuth` is `true`? As I said earlier, it should be the same as in the `other`.

madf (changes requested):

tests/packet_tests.cpp
```diff
@@ -0,0 +1,188 @@
...
+{
+       std::vector<uint8_t> d {0x02, 0xd0, 0x00, 0x4d, 0x93, 0xa9, 0x61, 0x8b, 0x2f, 0x4c, 0x5a, 0x51, 0x65, 0x67, 0x3d, 0xb4, 0x07, 0x30, 0xa2, 0x39, 0x01, 0x06, 0x74, 0x65, 0x73, 0x74, 0x05, 0x06, 0x00, 0x00, 0x00, 0x14, 0x04, 0x06, 0x7f, 0x68, 0x16, 0x11, 0x13, 0x08, 0x31, 0x32, 0x33, 0x61, 0x62, 0x63, 0x03, 0x13, 0x01, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x38, 0x39, 0x61, 0x62, 0x63, 0x64, 0x65, 0x66, 0x67, 0x1a, 0x0c, 0x00, 0x00, 0x00, 0xab, 0x01, 0x06, 0x00, 0x00, 0x00, 0x03};
+
+    RadProto::Packet p(d.data(), d.size(), "secret");
+
+    BOOST_CHECK_EQUAL(p.type(), 2);
+
+    BOOST_CHECK_EQUAL(p.id(), 0xd0);
+
+    std::array<uint8_t, 16> authValues = p.auth();
```
madf, line 126: You do not need this variable. You can use `p.auth()` directly.

src/socket.cpp
```diff
@@ -41,7 +41,7 @@ void Socket::asyncSend(const Packet& response, const std::function<void(const er
     const std::vector<uint8_t> vResponse = response.makeSendBuffer(m_secret);
     std::copy(vResponse.begin(), vResponse.end(), m_recvBuffer.begin());
```
madf, line 42: You use `m_recvBuffer` for both sending and receiving data. Let's call it just `m_buffer`.

tests/packet_tests.cpp
madf: Same in the previous test case.

madf (changes requested):

PR #49, opened 2024-08-08
Async Radius. Unit test.

HelenMamontova:
The new file dictionaries_tests.cpp with test case BasicDictionary added.

### DIFF
--- a/include/radproto/error.h
+++ b/include/radproto/error.h
@@ -29,7 +29,7 @@ namespace RadProto
     class Exception: public std::runtime_error
     {
         public:
-            explicit Exception(const boost::system::error_code& errorCode, std::string message);
+            explicit Exception(const boost::system::error_code& errorCode, const std::string& message);
             explicit Exception(const boost::system::error_code& errorCode);
             boost::system::error_code getErrorCode() const {return m_errorCode;}
 
--- a/include/radproto/error.h
+++ b/include/radproto/error.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <boost/system/error_code.hpp>
+#include <string>
 
 namespace RadProto
 {
@@ -20,12 +21,15 @@ namespace RadProto
         eapMessageAttributeError,
         invalidAttributeType,
         invalidAttributeSize,
-        invalidVendorSpecificAttributeId
+        invalidVendorSpecificAttributeId,
+        suchAttributeNameAlreadyExists,
+        suchAttributeCodeAlreadyExists
     };
 
     class Exception: public std::runtime_error
     {
         public:
+            explicit Exception(const boost::system::error_code& errorCode, std::string message);
             explicit Exception(const boost::system::error_code& errorCode);
             boost::system::error_code getErrorCode() const {return m_errorCode;}
 
--- a/src/dictionaries.cpp
+++ b/src/dictionaries.cpp
@@ -21,7 +21,7 @@ void BasicDictionary::add(uint32_t code, const std::string& name)
     for (const auto& entry: m_rightDict)
     {
         if (entry.second == name && entry.first != code)
-            throw RadProto::Exception(RadProto::Error::suchAttributeNameAlreadyExists, "BasicDictionary::add. Attribute name " + name + " already exists with code " + std::to_string(entry.first));
+            throw RadProto::Exception(RadProto::Error::suchAttributeNameAlreadyExists, "[BasicDictionary::add]. Attribute name " + name + " already exists with code " + std::to_string(entry.first));
         if (entry.second == name && entry.first == code)
             continue;
     }
@@ -36,7 +36,7 @@ void BasicDictionary::append(const BasicDictionary& basicDict)
         for (const auto& item: m_rightDict)
         {
             if (entry.second == item.second && entry.first != item.first)
-                throw RadProto::Exception(RadProto::Error::suchAttributeNameAlreadyExists, "BasicDictionary::append. Attribute name " + entry.second + " already exists with code " + std::to_string(item.first));
+                throw RadProto::Exception(RadProto::Error::suchAttributeNameAlreadyExists, "[BasicDictionary::append]. Attribute name " + entry.second + " already exists with code " + std::to_string(item.first));
             if (entry.second == item.second && entry.first == item.first)
                 continue;
         }
--- a/src/dictionaries.cpp
+++ b/src/dictionaries.cpp
@@ -19,11 +19,12 @@ uint32_t BasicDictionary::code(const std::string& name) const
 void BasicDictionary::add(uint32_t code, const std::string& name)
 {
     for (const auto& entry: m_rightDict)
+    {
         if (entry.second == name && entry.first != code)
             throw RadProto::Exception(RadProto::Error::suchAttributeNameAlreadyExists, "BasicDictionary::add. Attribute name " + name + " already exists with code " + std::to_string(entry.first));
-        else if (entry.second == name && entry.first == code)
+        if (entry.second == name && entry.first == code)
             continue;
-
+    }
     m_rightDict.insert_or_assign(code, name);
     m_reverseDict.emplace(name, code);
 }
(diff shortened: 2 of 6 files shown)


tests/dictionaries_tests.cpp
```diff
@@ -0,0 +1,44 @@
...
+    RadProto::BasicDictionary b;
+
+    BOOST_CHECK_THROW(b.name(0), std::out_of_range);
+
+    BOOST_CHECK_THROW(b.code(""), std::out_of_range);
+
+    b.add(1, "User-Name");
+    b.add(2, "abc");
+    b.add(2, "def");
+    b.append(b);
```
madf, line 30: Why do you append the dictionary to itself?

tests/dictionaries_tests.cpp
```diff
@@ -0,0 +1,44 @@
...
+{
+    RadProto::BasicDictionary b;
+
+    BOOST_CHECK_THROW(b.name(0), std::out_of_range);
+
+    BOOST_CHECK_THROW(b.code(""), std::out_of_range);
+
+    b.add(1, "User-Name");
+    b.add(2, "abc");
+    b.add(2, "def");
```
madf, line 29: I think it would be natural for this statement to override 'abc' with 'def'. So `b.name(2)` would return 'def' and `b.code("def")` would return 2.

It is unexpected that the dictionary can contain two entities with the same code.

tests/dictionaries_tests.cpp
```diff
@@ -0,0 +1,44 @@
...
+    b.append(b);
+
+    BOOST_CHECK_EQUAL(b.name(1), "User-Name");
+    BOOST_CHECK_EQUAL(b.name(2), "abc");
+
+    BOOST_CHECK_EQUAL(b.code("User-Name"), 1);
+    BOOST_CHECK_EQUAL(b.code("abc"), 2);
+    BOOST_CHECK_EQUAL(b.code("def"), 2);
+
+    BOOST_CHECK_EQUAL(b.rightDict().at(2), "abc");
```
madf, line 39: Please remind me why do we have `rightDict` and `reverseDict` methods in public? How do we use them?

madf (changes requested):

include/radproto/error.h
```diff
@@ -20,12 +21,15 @@ namespace RadProto
...
-        invalidVendorSpecificAttributeId
+        invalidVendorSpecificAttributeId,
+        suchAttributeNameAlreadyExists,
+        suchAttributeCodeAlreadyExists
     };
 
     class Exception: public std::runtime_error
     {
         public:
+            explicit Exception(const boost::system::error_code& errorCode, std::string message);
```
madf, line 32: `const std::string& message`

src/error.cpp
```diff
@@ -2,6 +2,12 @@
 #include <string>
 
 using Exception = RadProto::Exception;
+Exception::Exception(const boost::system::error_code& errorCode, std::string message)
+    : std::runtime_error(errorCode.message() + " - " + message),
```
madf, line 6: I guess, the meaning of this constructor is to completely substitute the autogenerated message. So not `errorCode.message() + " - " + message` but just `message`.

src/dictionaries.cpp
```diff
@@ -18,17 +18,27 @@ uint32_t BasicDictionary::code(const std::string& name) const
 
 void BasicDictionary::add(uint32_t code, const std::string& name)
 {
-    m_rightDict.emplace(code, name);
+    for (const auto& entry: m_rightDict)
+        if (entry.second == name)
+            throw RadProto::Exception(RadProto::Error::suchAttributeNameAlreadyExists, name + "- add");
```
madf, line 23: Taking into account my comment above, here you should provide the full message, not just a suffix.

madf: Also, what if we add the same name with the same code? As we discussed it earlier, this should be a no-op.

tests/dictionaries_tests.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+    BOOST_CHECK_THROW(b.name(0), std::out_of_range);
+
+    BOOST_CHECK_THROW(b.code(""), std::out_of_range);
+
+    b.add(1, "User-Name");
+    b.add(1, "User");
+    b.add(5, "ijk");
+    b.add(3, "def");
+    BOOST_CHECK_THROW(b.add(2, "User"), RadProto::Exception);
```
madf, line 132: Please also ensure that `b.add(1, "User")` does not throw.

tests/dictionaries_tests.cpp
```diff
@@ -0,0 +1,154 @@
...
+    RadProto::BasicDictionary c;
+    c.add(2, "def");
+    c.add(5, "ghi");
+
+    BOOST_CHECK_EQUAL(c.name(2), "def");
+    BOOST_CHECK_EQUAL(c.name(5), "ghi");
+    BOOST_CHECK_EQUAL(c.code("def"), 2);
+    BOOST_CHECK_EQUAL(c.code("ghi"), 5);
+
+    BOOST_CHECK_THROW(c.append(b), RadProto::Exception);
```
madf, line 152: Please also check that `append` does not throw if 'def' has a code 3.

madf (changes requested):

src/dictionaries.cpp
```diff
@@ -18,17 +18,31 @@ uint32_t BasicDictionary::code(const std::string& name) const
 
 void BasicDictionary::add(uint32_t code, const std::string& name)
 {
-    m_rightDict.emplace(code, name);
+    for (const auto& entry: m_rightDict)
+        if (entry.second == name && entry.first != code)
+            throw RadProto::Exception(RadProto::Error::suchAttributeNameAlreadyExists, "Function add. Attribute name " + name + " already exists with code " + std::to_string(entry.first));
```
madf, line 23: Instead of 'Function add' I would use something like '[BasicDictionary::add]'. First, it is more specific. Second, it clearly separates location information from the message itself.

Same below.

src/dictionaries.cpp
```diff
@@ -18,17 +18,31 @@ uint32_t BasicDictionary::code(const std::string& name) const
...
 }
 
 void BasicDictionary::append(const BasicDictionary& basicDict)
 {
     for (const auto& entry: basicDict.m_rightDict)
+    {
+        for (const auto& item: m_rightDict)
+            if (entry.second == item.second && entry.first != item.first)
+                throw RadProto::Exception(RadProto::Error::suchAttributeNameAlreadyExists, "Function append. Attribute name " + entry.second + " already exists with code " + std::to_string(item.first));
+            else if (entry.second == item.second && entry.first == item.first)
```
madf, line 38: You do not need `else` after `throw`.

tests/dictionaries_tests.cpp
```diff
@@ -0,0 +1,190 @@
...
+#pragma GCC diagnostic ignored "-Wsign-compare"
+#pragma GCC diagnostic ignored "-Wparentheses"
+#include <boost/test/unit_test.hpp>
+#pragma GCC diagnostic pop
+
+BOOST_AUTO_TEST_SUITE(DictionariesTests)
+
+BOOST_AUTO_TEST_SUITE(BasicDictionaryTests)
+
+BOOST_AUTO_TEST_CASE(FunctionAdd)
```
madf, line 22: I would call this test 'TestAdd'. 'FunctionAdd' in English does not mean "a function called 'add'", it means either addition of functions or something 'functional add'. OTOH, 'TestAdd' clearly says that here we test something called 'add'.

madf (changes requested):

src/dictionaries.cpp
```diff
@@ -19,11 +19,12 @@ uint32_t BasicDictionary::code(const std::string& name) const
 void BasicDictionary::add(uint32_t code, const std::string& name)
 {
     for (const auto& entry: m_rightDict)
+    {
         if (entry.second == name && entry.first != code)
-            throw RadProto::Exception(RadProto::Error::suchAttributeNameAlreadyExists, "Function add. Attribute name " + name + " already exists with code " + std::to_string(entry.first));
-        else if (entry.second == name && entry.first == code)
+            throw RadProto::Exception(RadProto::Error::suchAttributeNameAlreadyExists, "BasicDictionary::add. Attribute name " + name + " already exists with code " + std::to_string(entry.first));
```
madf, line 24: 'BasicDisctionary::add' -> '[BasicDictionary::add]'.

Same below.

madf (changes requested):

PR #54, opened 2025-02-19
Async Radius

HelenMamontova:
Method close() added to class Socket.

### DIFF
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -65,12 +65,8 @@ void Socket::handleSend(const error_code& ec, const std::function<void(const err
     callback(ec);
 }
 
-boost::system::error_code Socket::close(error_code& ec)
+void Socket::close(error_code& ec)
 {
-    if (m_socket.is_open())
-    {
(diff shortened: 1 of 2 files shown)


src/socket.cpp
```diff
@@ -64,3 +64,13 @@ void Socket::handleSend(const error_code& ec, const std::function<void(const err
 {
     callback(ec);
 }
+
+boost::system::error_code Socket::close(error_code& ec)
+{
+    if (m_socket.is_open())
```
madf, line 70: I do not think you need this check.

src/socket.cpp
```diff
@@ -64,3 +64,13 @@ void Socket::handleSend(const error_code& ec, const std::function<void(const err
...
+    {
+        m_socket.shutdown(udp::socket::shutdown_both, ec);
+        m_socket.close(ec);
+    }
+    return ec;
```
madf, line 75: `ec` is an out-param, no need to return it.

madf (changes requested):

PR #55, opened 2025-04-08
Make it possible to build async-radius as git submodule + cmake add_s…

yoori:
…ubdirectory in other project

Hi, great work ! Thank you !
Here is trivial changes, that allow to build your project inside other project source tree as git submodule + add_subdirectory ( without cmake FetchContent)


madf:
Should `PROJECT_SOURCE_DIR` do the same thing?

yoori:
Difference that CMAKE_SOURCE_DIR is root of full project, when we build asynchronous-radius inside other project (as subdirectory) it refers to non async-radius root …

yoori:
You right, we can use it and this is more consistent - I will change pull request

madf:
Ok. I'm waiting for the changes.

yoori:
Done

madf:
Merged. Thanks for contribution! I'll make a new release soon.

PR #64, opened 2025-12-12
Сhanges in class Attribute.

HelenMamontova:
The make function added to class Attribute.


madf (changes requested):

madf:
It would be nice to have tests for new function.

